PR #8, opened 2019-05-29
Created matrix to graph conversion function

pkj-m:
Added function called matrix2graph to convert input sparse matrix to a graph with vertices representing columns.
Time taken is proportional to the product of the square of the number of columns and number of rows.

### DIFF
--- a/src/matrix2graph.jl
+++ b/src/matrix2graph.jl
@@ -6,7 +6,7 @@ include("custom_graph.jl")
 A utility function to generate a graph from input
 sparse matrix, columns are represented with vertices
 and 2 vertices are connected with an edge only if
-the two columns are mutually orthogonal.
+the two columns are not mutually orthogonal.
 """
 function matrix2graph(SparseMatrix::Array{Int64,2})
     V = col = size(SparseMatrix, 2)
--- a/src/matrix2graph.jl
+++ b/src/matrix2graph.jl
@@ -1,0 +1,50 @@
+include("custom_graph.jl")
+
+"""
+        matrix2graph(SparseMatrix)
+
+A utility function to generate a graph from input
+sparse matrix, columns are represented with vertices
+and 2 vertices are connected with an edge only if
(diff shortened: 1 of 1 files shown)


src/matrix2graph.jl
```diff
@@ -0,0 +1,50 @@
...
+            if oneMatrix' * colRes > 0
+                add_edge!(graph, i, j)
+            end
+        end
+    end
+    return graph
+end
+
+function element_or(col1, col2)
+    c3 = zeros(Int64,0)
```
ChrisRackauckas: Use `Int` instead of `Int64`

src/matrix2graph.jl
```diff
@@ -0,0 +1,50 @@
...
+
+"""
+        matrix2graph(SparseMatrix)
+
+A utility function to generate a graph from input
+sparse matrix, columns are represented with vertices
+and 2 vertices are connected with an edge only if
+the two columns are not mutually orthogonal.
+"""
+function matrix2graph(SparseMatrix::Array{Int64,2})
```
ChrisRackauckas: You'll want this to specialize on `SparseMatrixCSC{T}`. In that, there is 

```julia
using SparseArrays
findnz(sprand(100,100,0.01))
```

to get the non-zero value locations

pkj-m: Used SparseArrays in the code instead of simple Arrays{Int64}. The inbuilt functions also allowed me to simplify the code a lot more without using all the logical comparisons that were being done in the previous version.

src/matrix2graph.jl
pkj-m: Issue doesn't exist anymore with the new code.

ChrisRackauckas:
Needs tests, but that's hard to do without LightGraphs.jl. I think the next step should be to change the custom graph type to use an AbstractGraph of LightGraphs.jl (or subtype the current one), and then write a test for this that uses graph isomorphism.

PR #9, opened 2019-06-01
sparsity of the jacobian of a program

shashi:
Given a program, enhances it to compute the sparsity of the Jacobian of that program by tracking which input indices are used in which output indices.

See examples at the end of the file.

If there are no branch cuts in the given function which result in different sparsity patterns, the sparsity can be computed once and reused, if not, we will need to run it on inputs that will take all the branch cuts and find the union of them. I'm looking for test cases with branch cuts that are not just `max` and `min`. :) 

It's possible that something like ConcolicFuzzer's approach may figure out inputs that will take all branches.

Thanks to @vchuravy for teaching me some Cassette and compilation stuff!

TODO:

- [ ] test coverage
- [ ] branch cuts
- [ ] benchmark compilation speed, try to improve. Using `broadcast` takes a long time!
- [ ] integrate the API with the rest of the DiffEq ecosystem.

### DIFF
--- a/src/program_sparsity.jl
+++ b/src/program_sparsity.jl
@@ -1,0 +1,203 @@
+using Cassette
+import Cassette: tag, untag, Tagged, metadata, hasmetadata, istagged
+
+"""
+The sparsity pattern.
+
+- `I`: Input index
+- `J`: Ouput index
+
+`(i, j)` means the `j`th element of the output depends on
+the `i`th element of the input. Therefore `length(I) == length(J)`
+"""
+struct Sparsity
+    m::Int
+    n::Int
+    I::Vector{Int} # Input
+    J::Vector{Int} # Output
+end
+
+using SparseArrays
+SparseArrays.sparse(s::Sparsity) = sparse(s.I, s.J, true, s.m, s.n)
+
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
Looks good. I'll merge it as is, refactor a bit, and open an issue to track developments. I prefer shorter 1-purpose PRs so they don't go stale, so this can keep getting improvements.

src/program_sparsity.jl
```diff
@@ -0,0 +1,203 @@
...
+"""
+function sparsity!(f!, Y, X, S=Sparsity(length(Y), length(X)))
+
+    ctx = SparsityContext(metadata=S)
+    ctx = Cassette.enabletagging(ctx, f!)
+    ctx = Cassette.disablehooks(ctx)
+
+    val = Cassette.overdub(ctx,
+                           f!,
+                           tag(Y, ctx, Output()),
```
ChrisRackauckas: should it `vec(Y)` here?

shashi: No I use `LinearIndexing` where needed, I feel like we shouldn't change the argument to `f!` itself.

PR #12, opened 2019-06-04
Replacing custom graph struct with LightGraphs abstraction

pkj-m:
Modified greedy_dist1 coloring and matrix2graph functions and tests to use [VertexSafeGraphs](https://github.com/matbesancon/VertexSafeGraphs.jl) as the default graph data type instead of custom_graph that was being used earlier.


test/test_contraction.jl
```diff
@@ -0,0 +1,36 @@
...
+    push!(test_graphs, copy(graph))
+end
+
+for i in 1:20
+    g = test_graphs[i]
+    out_colors = contract_color(g)
+    for v = 1:nv(g)
+        color = out_colors[v]
+        for j in inneighbors(g, v)
+            if out_colors[j] == color
```
matbesancon: the whole condition could be replaced by `@test out_color[j] != color`

PR #13, opened 2019-06-06
sparsity detection: a pass to try all tainted branches

shashi:
A branch is tainted if it depends on one of the inputs to the function `f`.

TODO:

- [x] Taint all args that are not `Input` or `Output`.
- [x] Allow a `Fixed` wrapper which says compute the sparsity while considering an argument to be fixed.

Demo:

```julia
julia> function foo(Y, X)
           if 1>X[1]
               Y[3, :] .= X
           else
               Y[2, :] .= X
           end
           Y
       end
foo (generic function with 1 method)

julia> sparsity!(foo, rand(3,3), [23,53,83])
Sparsity(9, 3, [2, 5, 8, 3, 6, 9], [1, 2, 3, 1, 2, 3])
```


ChrisRackauckas:
Walk me through this on Monday?

ChrisRackauckas:
Still WIP?

shashi:
Nope.  Just waiting for CI.

ChrisRackauckas:
Let's discuss tomorrow with @YingboMa and I in the lab. We should at least have 1 other person who knows how this works :).

Would there be a reason to keep the simpler pass as well? Speed? Compile time?

shashi:
Cool, I'll be in the lab tomorrow. Well the "simpler" pass is the default now, unless extra work is needed.

shashi:
Seems to work on Julia 1.2 !

PR #14, opened 2019-06-08
Struct for storing graph coloring

pkj-m:
Added a coloring struct to store and pass graph colors
as an argument more succinctly. 
Inspired from coloring in LightGraphs.jl

### DIFF
--- a/src/SparseDiffTools.jl
+++ b/src/SparseDiffTools.jl
@@ -2,7 +2,20 @@ module SparseDiffTools
 
 export  contract_color,
         greedy_d1,
-        matrix2graph
+        matrix2graph,
+        Coloring
+
+"""
+        Coloring{T}
+
+Stores the number of colors used in a graph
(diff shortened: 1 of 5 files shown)


ChrisRackauckas:
IMO we should really just use a Vector of Integers for this.

pkj-m:
I was thinking using a struct to store/pass the coloring might be a good idea since it'd reflect the data that we're storing in the array more accurately. Plus, the struct itself contains just that array plus an extra value storing the total number of unique colors used.

ChrisRackauckas:
>Plus, the struct itself contains just that array plus an extra value storing the total number of unique colors used.

Yes, but that simplicity is what makes it unnecessary. `color = 1:n` is easy, reflects a dense matrix, and is constant memory. `maximum(color)` does the job for counting colors. We shouldn't make something more complicated than it needs to be.

pkj-m:
Fair enough. Closing.

PR #15, opened 2019-06-10
[WIP] Finite Differencing along coloring vector

pkj-m:
This PR aims to tackle to issue of calculating the jacobian
matrix using finite difference method, particularly in a specific
chosen directions that are computed from the coloring of the
graph representation of the derivative matrix, which has been
done previously.

The function to calculate the direction vectors given the
coloring has been added, however, it will be integrated with
the main finite-diff function in the next commit.
Target list before merging:

1. - [ ] use direction vectors while computing derivative

2. - [ ] add examples demonstrating use

3. - [ ] add tests for relevant functions

### DIFF
--- a/src/finite_diff.jl
+++ b/src/finite_diff.jl
@@ -1,0 +1,63 @@
+"""
+    finite_diff(J, f, fx, coloring)
+
+Compute the jacobian of function  f  at  given
+point x via central forward differencing along
+direction vectors determined by graph coloring
+"""
+function finite_diff(
+    f,
+    x::AbstractArray{<:Number},
+    coloring::AbstractArray{<:Integer})
+
+    num_sets = maximum(coloring)
+
+    fx = f(x)
+    n = length(x)
+    m = length(fx)
+
+    J = zeros(Float64, m , n)
+
+    A = rand(n,n);
+    id = diagm(0=>fill(1., size(A,1)))
+
+    ep = sqrt(eps(real(Float64)))
+
+    for i = 1:n
+        dx = x + ep.*id[:,i]
+        J[:,i] = (f(dx)-fx) / ep
+    end
+
+    return J
+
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
Don't create a new finite difference function, just update the old one to use the color vectors.

src/finite_diff.jl
```diff
@@ -0,0 +1,63 @@
...
+
+
+"""
+        find_vectors(coloring)
+
+Calculate the direction vectors needed for forward
+difference function from the given coloring  of  a
+graph
+"""
+function find_vectors(coloring::AbstractArray{<:Integer})
```
ChrisRackauckas: You don't want to allocate like this

src/finite_diff.jl
```diff
@@ -0,0 +1,63 @@
...
+
+    J = zeros(Float64, m , n)
+
+    A = rand(n,n);
+    id = diagm(0=>fill(1., size(A,1)))
+
+    ep = sqrt(eps(real(Float64)))
+
+    for i = 1:n
+        dx = x + ep.*id[:,i]
```
ChrisRackauckas: We don't want to use an allocating form

ChrisRackauckas:
This is very much in the wrong direction so I am going to close it.

src/finite_diff.jl
pkj-m: Can you explain how to pass the coloring vector instead?

ChrisRackauckas: Just add a keyword argument for a vector of integers for the colors

ChrisRackauckas: Let me know if you want starter code

pkj-m: Yeah I think that'd be of great help at the moment.

PR #16, opened 2019-06-15
Added integration test

pkj-m:
1. Added integration test for testing full stack. 
2. Modified the matrix2graph function to accept sparse matrices with elements <: Number, not just integers
3. Added LinearAlgebra and DiffEqDiffTools as dependencies.


src/matrix2graph.jl
```diff
@@ -10,7 +10,7 @@ sparse matrix, columns are represented with vertices
 and 2 vertices are connected with an edge only if
 the two columns are mutually orthogonal.
 """
-function matrix2graph(SparseMatrix::SparseMatrixCSC{Int64,Int64})
+function matrix2graph(SparseMatrix::SparseMatrixCSC{T,Int64}) where T<:Number
```
ChrisRackauckas: Instead of `Int64`, use `Int`.

Project.toml
```diff
@@ -4,7 +4,9 @@ authors = ["Pankaj Mishra <pankajmishra1511@gmail.com>"]
 version = "0.1.0"
 
 [deps]
+DiffEqDiffTools = "01453d9d-ee7c-5054-8395-0335cb756afa"
```
ChrisRackauckas: This only needs to be a test dependency

pkj-m: Done. Moved to extras.

src/matrix2graph.jl
pkj-m: Changed to Int.

PR #17, opened 2019-06-19
Automatic Differentiation with coloring

pkj-m:
Added function `compute_jacobian!` to compute Jacobian matrices using a coloring vector. Also added tests.

### DIFF
--- a/src/compute_jacobian_ad.jl
+++ b/src/compute_jacobian_ad.jl
@@ -1,0 +1,40 @@
+using ForwardDiff: Dual, jacobian, partials
+
+function compute_jacobian!(J::AbstractMatrix{<:Number},
+                f,
+                x::AbstractArray{<:Number};
+                color=1:length(x))
+
+    partials_array = Array{Float64}(undef, length(x), maximum(color))
+    for color_i in 1:maximum(color)
+        for i in 1:length(x)
+            if color[i]==color_i
+                partials_array[i,color_i] = 1
+            else
+                partials_array[i,color_i] = 0
+            end
+        end
+    end
+
+    p = Tuple.(eachrow(partials_array))
+    t = zeros(Dual{Nothing, Float64, maximum(color)},0)
+    for i in 1:length(x)
+       push!(t, Dual(x[i], p[i]))
+    end
+
+    fx = similar(t)
+    f(fx, t)
+
+    rows_index, cols_index, val = findnz(J)
+    for color_i in 1:maximum(color)
+        du = partials.(fx,color_i)
+        for i in 1:length(cols_index)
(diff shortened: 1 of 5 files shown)


test/test_ad.jl
```diff
@@ -0,0 +1,37 @@
+using SparseDiffTools
+using SparseArrays
+using ForwardDiff: Dual, jacobian
+
+fcalls = 0
+function f(dx,x)
```
matbesancon: Prefer 4 spaces for indentation

src/compute_jacobian_ad.jl
```diff
@@ -0,0 +1,40 @@
...
+            if color[i]==color_i
+                partials_array[i,color_i] = 1
+            else
+                partials_array[i,color_i] = 0
+            end
+        end
+    end
+
+    p = Tuple.(eachrow(partials_array))
+    t = zeros(Dual{Nothing, Float64, maximum(color)},0)
```
ChrisRackauckas:
```suggestion
    t = zeros(Dual{typeof(f), Float64, maximum(color)},length(x))
```

It's usually better to pre-allocate. The first part here being `typeof(f)` will tag the Duals so they won't accidentally overlap with other differentiations.

src/compute_jacobian_ad.jl
```diff
@@ -0,0 +1,40 @@
...
+    for i in 1:length(x)
+       push!(t, Dual(x[i], p[i]))
```
ChrisRackauckas:
```suggestion
       t[i] = Dual(x[i], p[i])
```

src/compute_jacobian_ad.jl
```diff
@@ -0,0 +1,40 @@
...
+    for color_i in 1:maximum(color)
+        du = partials.(fx,color_i)
+        for i in 1:length(cols_index)
+            if color[cols_index[i]]==color_i
+                J[rows_index[i],cols_index[i]] = du[rows_index[i]]
+            end
+        end
+    end
+
+    J = Array(J)
```
ChrisRackauckas: This isn't needed.

ChrisRackauckas:
So there's quite a few things that can be improved performance-wise, but it makes sense to do that in a follow up PR. Ideally we would just allocate everything upfront, and let the user pass in a cache that will store everything. Then we just need to use those cache variables to effectively do the computation without allocating. I'll merge and at least do a partial PR to show what I mean by this.

Additionally, we should be using chunk sizes. I find https://github.com/FluxML/Tracker.jl/blob/master/src/forward.jl a good reference on how to implement that. That will be easier to implement when we have a cache.

So I'm going to merge and we can keep handling this. I'll fix the tab spacing because that was my doing.

Issue #23, opened 2019-06-22
Respect the chunksize in forward-mode AD


ChrisRackauckas:
https://github.com/JuliaDiffEq/SparseDiffTools.jl/blob/master/src/differentiation/compute_jacobian_ad.jl#L47

ChrisRackauckas:
https://github.com/FluxML/Zygote.jl/blob/ce8111da692f776f7a8973cf32497319252a0559/src/lib/forward.jl#L12-L34 shows how to do it.

pkj-m:
Consider this Jacobian matrix :

![Screenshot from 2019-05-28 17-23-15](https://user-images.githubusercontent.com/17577971/59983307-aeb75c80-963b-11e9-9c42-c7e93358afe2.png)

This will give us the coloring vector: `[1,2,3,1,3]`
which in turn will give us the seed matrix (from which we derive the partials):

~~~
[[1, 0, 0],
 [0 ,1 ,0],
 [0, 0, 1],
 [1, 0, 0],
 [0, 0, 1]]
~~~
Now say if the chunksize is set to 3, then we are supposed to pass only 3 partials for the first AD call, so the partials that we'll be passing would look something like:
~~~
[[1, 0, 0],
 [0 ,1 ,0],
 [0, 0, 1],
 [0, 0, 0],
 [0, 0, 0]]
~~~
My idea is, instead of immediately passing these partials to the `forwarddiff_color_jacobian!`, can we store them in an array, say `partials_array`, whose first element will be the above shown partials list (converted to a tuple perhaps), and second element would look like:

~~~
[[0, 0, 0],
 [0 ,0 ,0],
 [0, 0, 0],
 [1, 0, 0],
 [0, 0, 1]]
~~~
Now we can pass this vector of partials to the function, iterate over it in the function body, for each `partial_i in partials_array`, compute the partial Jacobian, and save it in `J`. We don't decompress it yet. Instead, as we move to the next value of `partial_i` and compute more values of the Jacobian, we add them to the previously computed `J` itself. There will be no collision due to the way coloring works. Once all the partials have been computed, our compressed Jacobian will consist of all the final values. Now we can use the previously implemented decompression logic to obtain the original Jacobian.

Can you tell me if you see any flaws or problems with the logic or the implementation anywhere? I think you probably wanted me to decompress the column elements into their actual positions with each partial computation, but I believe this method would be much easier to implement without any obvious drawbacks or performance loss. Thoughts?

[...]

PR #35, opened 2019-06-25
Limit size of partials by chunk size

pkj-m:
Allows user to control the number of partials to be computed simultaneously by the program by passing the limit as chunk size.

Reference: #23

### DIFF
--- a/src/differentiation/compute_jacobian_ad.jl
+++ b/src/differentiation/compute_jacobian_ad.jl
@@ -1,4 +1,3 @@
-using SparseArrays
 struct ForwardColorJacCache{T,T2,T3,T4,T5}
     t::T
     fx::T2
@@ -123,27 +122,5 @@ function forwarddiff_color_jacobian!(J::AbstractMatrix{<:Number},
         end
 
     end
-
     nothing
-    #=
-    t .= Dual{typeof(f)}.(x, p)
-    f(fx, t)
-
-    if J isa SparseMatrixCSC
-        rows_index, cols_index, val = findnz(J)
-        for color_i in 1:maximum(color)
-            dx .= partials.(fx,color_i)
-            for i in 1:length(cols_index)
-                if color[cols_index[i]]==color_i
-                    J[rows_index[i],cols_index[i]] = dx[rows_index[i]]
-                end
-            end
-        end
-    else # Compute the compressed version
-        for color_i in 1:maximum(color)
-            J[:,i] .= partials.(fx,color_i)
-        end
-    end
-    nothing
-    =#
 end
--- a/src/differentiation/compute_jacobian_ad.jl
+++ b/src/differentiation/compute_jacobian_ad.jl
@@ -55,11 +55,7 @@ function generate_chunked_partials(x,color,::Val{N}) where N
 
     for color_i in 1:maximum(color)
         for j in 1:length(x)
-            if color[j]==color_i
-                partials[j,color_i] = true
-            else
-                partials[j,color_i] = false
-            end
+            partials[j, color_i] = color[j] == color_i
         end
     end
 
--- a/src/differentiation/compute_jacobian_ad.jl
+++ b/src/differentiation/compute_jacobian_ad.jl
@@ -128,6 +128,7 @@ function forwarddiff_color_jacobian!(J::AbstractMatrix{<:Number},
                     color_i = 1
                 end
             end
+        end
     end
     nothing
 end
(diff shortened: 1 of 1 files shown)


src/differentiation/compute_jacobian_ad.jl
```diff
@@ -44,20 +44,38 @@ end
...
+    chunked_partials = Array{Array{NTuple,1},1}(undef, num_of_chunks)
 
-    partials_array = BitMatrix(undef, length(x), maximum(color))
     for color_i in 1:maximum(color)
-        for i in eachindex(x)
-            if color[i]==color_i
-                partials_array[i,color_i] = true
+        for j in 1:length(x)
+            if color[j]==color_i
+                partials[j,color_i] = true
```
YingboMa: `partials[j,color_i] = color[j] == color_i`

src/differentiation/compute_jacobian_ad.jl
```diff
@@ -44,20 +44,34 @@ end
 generate_chunked_partials(x,color,N::Integer) = generate_chunked_partials(x,color,Val(N))
 function generate_chunked_partials(x,color,::Val{N}) where N
 
-    # TODO: should only go up to the chunksize each time, and should
-    # generate p[i] different parts, each with less than the chunksize
+    chunksize = getsize(default_chunk_size(maximum(color)))
```
ChrisRackauckas: we still aren't making use of the actual chunksize if passed by the user, so we should fix that up in a follow up.

src/differentiation/compute_jacobian_ad.jl
```diff
@@ -79,23 +93,37 @@ function forwarddiff_color_jacobian!(J::AbstractMatrix{<:Number},
...
+                end
+                color_i += 1
+                if color_i == maximum(color) + 1
+                    color_i = 1
+                end
+            end
+        else
+            for j in 1:chunksize
+                col_index = (i-1)*chunksize + j
+                J[:, col_index] .= partials(fx, color_i)
```
ChrisRackauckas: partials.(

ChrisRackauckas:
```suggestion
                J[:, col_index] .= partials.(fx, color_i)
```

src/differentiation/compute_jacobian_ad.jl
```diff
@@ -79,23 +93,37 @@ function forwarddiff_color_jacobian!(J::AbstractMatrix{<:Number},
...
+                color_i += 1
+                if color_i == maximum(color) + 1
```
ChrisRackauckas: this will only happen at the end?

ChrisRackauckas: Oh I see what this is. The name is confusing because it's not longer necessarily the color. Name it like `cur_partial`. And shouldn't it reset after it hits the chunksize? Otherwise `partials(fx, color_i)` can have `color_i > chunksize` and it will index a partial that doesn't exist

src/differentiation/compute_jacobian_ad.jl
```diff
@@ -79,23 +93,37 @@ function forwarddiff_color_jacobian!(J::AbstractMatrix{<:Number},
...
+            for j in 1:chunksize
+                dx .= partials.(fx, j)
+                for k in 1:length(cols_index)
+                    if color[cols_index[k]] == color_i
+                        J[rows_index[k], cols_index[k]] = dx[rows_index[k]]
+                    end
+                end
+                color_i += 1
+                if color_i == maximum(color) + 1
+                    color_i = 1
```
ChrisRackauckas: this doesn't seem to be used?

pkj-m: That's because the chunk size defined here 
~~~
function default_chunk_size(maxcolor)
    if maxcolor < DEFAULT_CHUNK_THRESHOLD
        Val(maxcolor)
    else
        Val(DEFAULT_CHUNK_THRESHOLD)
    end
end
~~~
ensures that the size of a chunk is always less than or equal to the number of colors used. If we decide to use chunk sizes that are larger than the maximum value of color, we'll have to make sure `color_i` wraps around instead of going past the largest value in coloring vector.

pkj-m: Oh okay nevermind my bad. That made no sense.

ChrisRackauckas:
Looks good! Still a few things I would change around, so I'll just merge and then send a PR when my flight is done.

src/differentiation/compute_jacobian_ad.jl
pkj-m: Yeah you're right. But I think we can just remove the `color_i` variable entirely and simply copy the columns from the partials to the Jacobian using
~~~
for j in 1:chunksize
    col_index = (i-1)*chunksize + j
    J[:, col_index] .= partials.(fx, j)
end
~~~
We use `col_index` variable to map which column in the original matrix would the the `jth` column in the `ith` pass of the chunked partials would correspond to, using the equation supplied.

PR #38, opened 2019-06-27
Vectorized decompression

ChrisRackauckas:
It is slower but potentially useful for GPUs:

```julia
# Vectorized: 13.366 μs (16 allocations: 5.08 KiB)
# Loop: 2.513 μs (4 allocations: 2.41 KiB)
```


matbesancon:
can't we specialize for GPU and keep the looping default?

ChrisRackauckas:
Yes, this is just prototyping. I wrote this on a plane and pushed it so we wouldn't lose what the correct translation is, because it's pretty gnarly. I may finish it up later or if @pkj-m has a GPU and wants to try pushing the GPU algorithm until `CuArrays.allowscalar(false)` works he can.

src/differentiation/compute_jacobian_ad.jl
```diff
@@ -98,6 +98,7 @@ function forwarddiff_color_jacobian!(J::AbstractMatrix{<:Number},
     sparsity = jac_cache.sparsity
     color_i = 1
     chunksize = length(first(first(jac_cache.p)))
+    fill!(J,zero(eltype(J)))
```
matbesancon:
```suggestion
    fill!(J, zero(eltype(J)))
```

matbesancon: Could also be a bit shorter with the eltype in the signature, with a `where` clause

src/differentiation/compute_jacobian_ad.jl
```diff
@@ -98,6 +98,7 @@ function forwarddiff_color_jacobian!(J::AbstractMatrix{<:Number},
...
 
     for i in 1:length(p)
```
matbesancon:
```suggestion
    for i in eachindex(p)
```

src/differentiation/compute_jacobian_ad.jl
```diff
@@ -107,9 +108,27 @@ function forwarddiff_color_jacobian!(J::AbstractMatrix{<:Number},
             rows_index, cols_index, val = findnz(sparsity)
             for j in 1:chunksize
                 dx .= partials.(fx, j)
-                for k in 1:length(cols_index)
-                    if color[cols_index[k]] == color_i
-                        J[rows_index[k], cols_index[k]] = dx[rows_index[k]]
+                if ArrayInterface.fast_scalar_indexing(x1)
+                    for i in 1:length(cols_index)
```
matbesancon: this i is shadowing the first, could we give another index name?

matbesancon: also, `eachindex`

src/differentiation/compute_jacobian_ad.jl
```diff
@@ -98,18 +98,37 @@ function forwarddiff_color_jacobian!(J::AbstractMatrix{<:Number},
...
+                        end
+                    end
+                else
+                    #=
+                    J.nzval[rows_index] .+= (color[cols_index] .== color_i) .* dx[rows_index]
+                    or
+                    J[rows_index, cols_index] .+= (color[cols_index] .== color_i) .* dx[rows_index]
+                    += means requires a zero'd out start
+                    =#
+                    if J isa SparseMatrixCSC
```
matbesancon: These are tricky to read, can we use the sugared version of `get/set index`?

ChrisRackauckas: Yeah, it's really tricky so I left the comment above. What's the sugared version of get/set index?

matbesancon: Ah ok I just realized there is no `J.nzval.[idxs] = ...` or so, with a dotted square bracket notation.

> What's the sugared version of get/set index?

`J[i] = J[y]`

ChrisRackauckas: yeah 🤷‍♂ https://github.com/JuliaLang/julia/issues/30845

src/coloring/high_level.jl
```diff
@@ -1,9 +1,9 @@
-abstract type ColoringAlgorithm end
-struct GreedyD1Color <: ColoringAlgorithm end
-struct BacktrackingColor <: ColoringAlgorithm end
-struct ContractionColor <: ColoringAlgorithm end
-struct GreedyStar1Color <: ColoringAlgorithm end
-struct GreedyStar2Color <: ColoringAlgorithm end
+abstract type SparseDiffToolsColoringAlgorithm <: ArrayInterface.ColoringAlgorithm end
```
matbesancon: `SparseDiffToolsColoringAlgorithm` not a fan of the package name plugged into types, C-like. Couldn't we not-export `ColoringAlgorithm`, thus calling it with `SparseDiffToolsColoringAlgorithm`. Other solution would be to remove the abstract type all together and use `ArrayInterface.ColoringAlgorithm` directly

test/test_ad.jl
```diff
@@ -56,3 +57,7 @@ jac_cache = ForwardColorJacCache(f,x,color = repeat(1:3,10), sparsity = _J1)
 forwarddiff_color_jacobian!(_denseJ1, f, x, jac_cache)
 @test _denseJ1 ≈ J
 @test fcalls == 1
+
+_Jt = similar(Tridiagonal(J))
+forwarddiff_color_jacobian!(_Jt, f, x, color = repeat(1:3,10), sparsity = _Jt)
+@test _Jt ≈ J
```
matbesancon:
```suggestion
@test _Jt ≈ J

```

PR #39, opened 2019-06-27
WIP: GPU-compatibility of Color Jacobians

ChrisRackauckas:
Needs to use Adapt.jl and easier ways to get CuSparseMatrixCSC

### DIFF
--- a/src/SparseDiffTools.jl
+++ b/src/SparseDiffTools.jl
@@ -12,6 +12,7 @@ using LinearAlgebra
 using SparseArrays, ArrayInterface
 
 using BlockBandedMatrices: blocksize, nblocks
+using CuArrays
 using ForwardDiff: Dual, jacobian, partials, DEFAULT_CHUNK_THRESHOLD
 
 using ArrayInterface: matrix_colors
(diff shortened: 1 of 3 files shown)


src/SparseDiffTools.jl
```diff
@@ -1,7 +1,8 @@
 module SparseDiffTools
 
 using SparseArrays, LinearAlgebra, BandedMatrices, BlockBandedMatrices,
-      LightGraphs, VertexSafeGraphs, DiffEqDiffTools, ForwardDiff, Zygote
+      LightGraphs, VertexSafeGraphs, DiffEqDiffTools, ForwardDiff, Zygote,
```
matbesancon: that makes a lot of packages required, could some of them be `import`-ed to simplify reading (or `using` only the identifiers needed)

test/test_ad.jl
```diff
@@ -40,3 +40,18 @@ jac_cache = ForwardColorJacCache(f,x,color = repeat(1:3,10))
 forwarddiff_color_jacobian!(_J1, f, x, jac_cache)
 @test _J1 ≈ J
 @test fcalls == 1
+
+using BenchmarkTools
```
matbesancon: a benchmark folder could be better to avoid putting this on CI every time

ChrisRackauckas: This isn't a real PR.

Issue #24, opened 2019-06-22
Jacobian row partitioning and reverse-mode AD


ChrisRackauckas:
Reverse-mode AD implementations calculate a Jacobian row by row instead of column by column. Thus it would be nice to have a way to do row-wise matrix partitioning and coloring. Once we have these color vectors, we have to specialize the Jacobian function in Zygote to allow for and exploit a color vector.

pkj-m:
Are we supposed to write the function for reverse mode AD from scratch like we did for forward mode or modify some pre-existing files in Zygote?

ChrisRackauckas:
Let's at least prototype in this repo, and then think about upstreaming it.

ChrisRackauckas:
You can use https://github.com/FluxML/Tracker.jl/blob/master/src/back.jl#L169-L178 as a template.

gdalle:
Apologies in advance for the repeated notifications, I'm going through every SparseDiffTools issue to see which ones are solved by SparseConnectivityTracer + SparseMatrixColorings + DifferentiationInterface.

This one is supported out of the box: if the user chooses a reverse mode backend, the coloring and compressed differentiation will automatically happen row-wise.

PR #40, opened 2019-06-28
Partition matrix by rows

pkj-m:
Added bool kwarg `partition_by_rows` which decides whether the partitioning of the matrix is done on the basis of rows or columns. Default value is false.
The color vector returned will be used for reverse-mode AD.

ref #24

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,8 +101,8 @@ of the Jacobian. The second argument is the optional choice of coloring algorith
 It will default to a greedy distance 1 coloring, though if your special matrix
 type has more information, like is a `Tridiagonal` or `BlockBandedMatrix`, the
 color vector will be analytically calculated instead. The variable argument
-partition_by_rows allows you to partition the Jacobian on the basis of rows instead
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -93,16 +93,19 @@ this can be significant savings.
...
-matrix_colors(A::AbstractMatrix,alg::ColoringAlgorithm = GreedyD1Color())
+matrix_colors(A::AbstractMatrix,alg::ColoringAlgorithm = GreedyD1Color(); partition_by_rows::Bool = false)
 ```
 
 The first argument is the abstract matrix which represents the sparsity pattern
 of the Jacobian. The second argument is the optional choice of coloring algorithm.
 It will default to a greedy distance 1 coloring, though if your special matrix
 type has more information, like is a `Tridiagonal` or `BlockBandedMatrix`, the
-color vector will be analytically calculated instead.
+color vector will be analytically calculated instead. The variable argument
```
ChrisRackauckas:
```suggestion
color vector will be analytically calculated instead. The keyword argument
```

PR #41, opened 2019-06-30
Refactor and update README to include sparsity detection

ChrisRackauckas:


src/program_sparsity/tuple.jl
```diff
@@ -1,36 +0,0 @@
...
-#
-# Tuple construction should preserve the tags of each component
-#
-
-@inline function Cassette.overdub(ctx::TaggedSparsity,
-                                  ::Core.Typeof(Core.tuple), args...)
-    tagged_new_tuple(ctx, args...)
-end
-
-# TODO: figure out why this is necessary and the previous method isn't enough
```
ChrisRackauckas: This file was never included @shashi haha

Issue #29, opened 2019-06-22
Coloring algorithms to be implemented


pkj-m:
- [ ] Greedy Star Coloring,  "[What Color is your Jacobian](https://pdfs.semanticscholar.org/ae4d/65769b6551a51d6fc6be2f021515bffa0798.pdf)?" pg 662
- [ ] Greedy Star Coloring 2,  "[What Color is your Jacobian](https://pdfs.semanticscholar.org/ae4d/65769b6551a51d6fc6be2f021515bffa0798.pdf)?" pg 663
- [ ] Backtracking Sequential Coloring (NP-hard),  "[Graph Coloring Algorithms](https://www.researchgate.net/publication/292100744_Graph_coloring_algorithms)"
- [ ] Star Bicoloring, "[What Color is your Jacobian](https://pdfs.semanticscholar.org/ae4d/65769b6551a51d6fc6be2f021515bffa0798.pdf)?" pg 668

ChrisRackauckas:
For Hessians, is it just the same coloring but on the adjacency matrix?

ChrisRackauckas:
What methods are used in the bidirectional coloring?

pkj-m:
In case of direct recovery, star coloring would work for Hessians. If we plan on extending the stack to include substitution based evaluation, we will have to employ acylic coloring as well.

pkj-m:
We can use either greedy distance-2 coloring (similar to distance-1 coloring) or partial distance-2 coloring in cases where we do not mix both forward and backward AD modes.

pkj-m:
Star coloring must satisfy the two conditions: (1) every pair
of adjacent vertices receives distinct colors (a distance-1 coloring), and (2) every path
on four vertices uses at least three colors. 

Acyclic coloring has the conditions: (1) the coloring corresponds to a distance-1 coloring, and (2) vertices in every cycle of the graph are assigned at least three distinct colors. The name acyclic comes from the fact that subgraph induced by vertices assigned any two colors is a collection of trees—and hence is acyclic.

reference: "[What Color is your Jacobian?]"(https://pdfs.semanticscholar.org/ae4d/65769b6551a51d6fc6be2f021515bffa0798.pdf)

ChrisRackauckas:
Add some acyclic coloring to the list as well then since substitution seems to be the proper way to handle bidirectional and forward-over-reverse.

pkj-m:
[...]

PR #43, opened 2019-07-01
New coloring methods

pkj-m:
- Implemented greedy star-1 coloring (ref: "[What Color is your Jacobian](https://pdfs.semanticscholar.org/ae4d/65769b6551a51d6fc6be2f021515bffa0798.pdf)?" pg 662 )
- Implemented greedy star-2 coloring (ref: "[What Color is your Jacobian](https://pdfs.semanticscholar.org/ae4d/65769b6551a51d6fc6be2f021515bffa0798.pdf)?" pg 663)

#29


pkj-m:
While it doesn't make a lot of sense to use star coloring for simple Jacobian computation using orthogonal column partitioning followed by forward mode AD or finite differencing, it can be used later for easier implementation of substitution-based recovery methods. 

I ran the new algorithm on some random and special graphs and this is what I observed.

For a star graph with 200,000 vertices:

-  the original greedy algorithm takes `46.170148 seconds (400.01 k allocations: 298.040 GiB, 19.11% gc time)`

- the new star coloring algorithm takes ` 98.273592 seconds (8 allocations: 3.052 MiB)` which is probably the worst case for this algorithm. 

So evidently the simple coloring algorithm performs better in this case, but not by an order or anything. We should also take note of the huge difference in the number of allocations and memory consumed.

For a cycle graph with again 200,000 vertices, 

- the original greedy algorithm takes ` 44.821616 seconds (400.01 k allocations: 298.040 GiB, 18.88% gc time)`

- the star coloring algorithm takes ` 0.008580 seconds (8 allocations: 3.052 MiB)`

Thus, for cycle graphs, our new star coloring algorithm is much, much faster than the original greedy coloring, in terms of both time and memory use. However, note that star coloring would return the coloring vector `[1, 2, 3, 1, 2, 3,...]` while the original algorithm would return `[1, 2, 1, 2...]`. Although I think the huge difference in time taken to compute the vector would easily nullify any advantage that one less `f` call could provide.

src/coloring/greedy_d1_coloring.jl
```diff
@@ -7,7 +7,7 @@ color using greedy approach. The number of colors
 used may be equal or greater than the chromatic
 number χ(G) of the graph.
 """
-function color_graph(G::VSafeGraph,alg::GreedyD1Color)
+function color_graph(G::VSafeGraph, alg::GreedyD1Color)
     V = nv(G)
```
matbesancon: prefer small letters for values (`g` for your graph, `v` for your number of vertices)

src/coloring/greedy_d1_coloring.jl
```diff
@@ -7,7 +7,7 @@ color using greedy approach. The number of colors
...
     result = zeros(Int64, V)
     result[1] = 1
```
matbesancon: below it seems you are only using the values 0 and 1 of available, you could replace it with a `BitVector`

src/coloring/greedy_star1_coloring.jl
```diff
@@ -0,0 +1,67 @@
+"""
+        greedy_star1_coloring
+
+    Find a coloring of a given input graph such that
+    no two vertices connected by an edge have the same
+    color using greedy approach. The number of colors
+    used may be equal or greater than the chromatic
+    number χ(G) of the graph.
```
matbesancon:
```suggestion
    number `χ(g)` of the graph.
```

src/coloring/greedy_star1_coloring.jl
```diff
@@ -0,0 +1,67 @@
...
+
+    1. every pair of adjacent vertices receives distinct  colors
+    (a distance-1 coloring)
+
+    2. For any vertex v, any color that leads to a two-colored path
+    involving v and three other vertices  is  impermissible  for  v.
+    In other words, every path on four vertices uses at least three
+    colors.
+
+    reference: What Color is your Jacobian?, pg 662
```
matbesancon:
```suggestion
    Reference: Gebremedhin AH, Manne F, Pothen A. **What color is your Jacobian? Graph coloring for computing derivatives.** SIAM review. 2005;47(4):629-705.
```

src/coloring/greedy_star1_coloring.jl
```diff
@@ -0,0 +1,67 @@
...
+"""
+function greedy_star1_coloring(G::VSafeGraph)
+    V = nv(G)
```
matbesancon: Same comment as above (small letters)

src/coloring/greedy_star1_coloring.jl
```diff
@@ -0,0 +1,67 @@
...
+    (a distance-1 coloring)
+
+    2. For any vertex v, any color that leads to a two-colored path
+    involving v and three other vertices  is  impermissible  for  v.
+    In other words, every path on four vertices uses at least three
+    colors.
+
+    reference: What Color is your Jacobian?, pg 662
+"""
+function greedy_star1_coloring(G::VSafeGraph)
```
matbesancon: given that you are not using the vertex-safety property, we could relax the type constraint to any graph, `g::LightGraphs.AbstractGraph`

src/coloring/greedy_star1_coloring.jl
```diff
@@ -0,0 +1,67 @@
...
+    V = nv(G)
+    color = zeros(Int64, V)
+
+    forbiddenColors = zeros(Int64, V+1)
+
+    for vertex_i = 1:V
```
matbesancon:
```suggestion
    for vertex_i = vertices(g)
```

src/coloring/greedy_star2_coloring.jl
```diff
@@ -0,0 +1,65 @@
...
+
+    1. every pair of adjacent vertices receives distinct  colors
+    (a distance-1 coloring)
+
+    2. For any vertex v, any color that leads to a two-colored path
+    involving v and three other vertices  is  impermissible  for  v.
+    In other words, every path on four vertices uses at least three
+    colors.
+
+    reference: What Color is your Jacobian?, pg 663
```
matbesancon: Same reference as above

src/coloring/greedy_star2_coloring.jl
```diff
@@ -0,0 +1,65 @@
+"""
+        greedy_star2_coloring
+
+    Find a coloring of a given input graph such that
+    no two vertices connected by an edge have the same
+    color using greedy approach. The number of colors
+    used may be equal or greater than the chromatic
+    number χ(G) of the graph.
```
matbesancon:
```suggestion
    number `χ(G)` of the graph.
```

src/coloring/greedy_star2_coloring.jl
```diff
@@ -0,0 +1,65 @@
...
+            end
+        end
+
+        color[vertex_i] = find_min_color(forbiddenColors, vertex_i)
+    end
+
+    color
+end
+
+function find_min_color(forbiddenColors::AbstractVector, vertex_i::Integer)
```
matbesancon: seems you have this function implemented twice, one can be removed

matbesancon:
Given that you don't use vertex removal, any graph can be passed (not only VSafeGraph). For the tests, generate a bunch of graphs, and compute the coloring using different algorithms. One basic property you can test is that the coloring is a valid one. Another one is that the number of colors is upper-bounded (you have to know the true minimum number of colors for that)

src/coloring/greedy_d1_coloring.jl
pkj-m: Done

src/coloring/greedy_d1_coloring.jl
```diff
@@ -1,30 +1,30 @@
...
-    V = nv(G)
-    result = zeros(Int64, V)
+function color_graph(g::VSafeGraph, alg::GreedyD1Color)
+    v = nv(g)
+    result = zeros(Int64, v)
     result[1] = 1
-    available = zeros(Int64, V)
-    for i = 2:V
-        for j in inneighbors(G, i)
+    available = BitArray{1}(undef, v)
```
matbesancon:
```suggestion
    available = BitVector(undef, v)
```

src/coloring/greedy_star1_coloring.jl
```diff
@@ -0,0 +1,67 @@
...
+            end
+        end
+
+        color[vertex_i] = find_min_color(forbiddenColors, vertex_i)
+    end
+
+    color
+end
+
+function find_min_color(forbiddenColors::AbstractVector, vertex_i::Integer)
```
matbesancon: prefer snake_case: `forbiddenColors`, same in the function above

test/test_greedy_star.jl
```diff
@@ -0,0 +1,83 @@
...
+        end
+        add_edge!(graph, v1, v2)
+    end
+    push!(test_graphs, copy(graph))
+end
+
+#=
+ Coloring needs to satisfy two conditions:
+
+1. every pair of adjacent vertices receives distinct  colors
```
matbesancon:
```suggestion
1. every pair of adjacent vertices receives distinct colors
```

test/test_greedy_star.jl
```diff
@@ -0,0 +1,83 @@
...
+    for v = vertices(g)
+        color = out_colors1[v]
+        for j in inneighbors(g, v)
+            @test out_colors1[j] != color
+        end
+    end
+
+    #test condition 2
+    for j = vertices(g)
+        walk = saw(g, j, 4)
```
matbesancon: Given the weird and ambiguous name of the function, it might be better to make it explicit
```suggestion
        walk = LightGraphs.saw(g, j, 4)
```

test/test_greedy_star.jl
```diff
@@ -0,0 +1,83 @@
...
+    for v = vertices(g)
+        color = out_colors2[v]
+        for j in inneighbors(g, v)
+            @test out_colors2[j] != color
+        end
+    end
+
+    #test condition 2
+    for j = vertices(g)
+        walk = saw(g, j, 4)
```
matbesancon:
```suggestion
        walk = LightGraphs.saw(g, j, 4)
```

matbesancon:
In your tests, you could add the two reference graphs from page 664, fig 4.3

matbesancon:
@pkj-m you are missing random in the package it seems, go to the project, `]add Random`

PR #47, opened 2019-07-11
add sparsity to forwarddiff_color_jacobian! and ForwardColorJacCache

huanglangwen:


src/differentiation/compute_jacobian_ad.jl
```diff
@@ -78,8 +80,9 @@ function forwarddiff_color_jacobian!(J::AbstractMatrix{<:Number},
                 f,
                 x::AbstractArray{<:Number};
                 dx = nothing,
-                color = eachindex(x))
-    forwarddiff_color_jacobian!(J,f,x,ForwardColorJacCache(f,x,dx=dx,color=color))
+                color = eachindex(x),
+                sparsity = nothing)
```
ChrisRackauckas: same here about the default

PR #48, opened 2019-07-12
Refactor sparsity detection

shashi:
1. Removes the `Path` type, uses `@amb` to execute branches non-deterministically
2. Better factoring of the tainted branches pass -- only API is `abstract_run`
3. `propagate_tags` framework to easily propagate known tags
4. Jacobian and Hessian-related code are now contained in jacobian.jl and hessian.jl only.

- [ ] This PR needs the Amb.jl package to be released first.


ChrisRackauckas:
This should be moved over to SparsityDetection.jl

matbesancon:
Should this be closed then?

PR #49, opened 2019-07-16
use Requires to suppress Cassette overwrite warnings

shashi:


ChrisRackauckas:
```
julia> using OrdinaryDiffEq
[ Info: Precompiling OrdinaryDiffEq [1dea7af3-3e70-54e6-95c3-0bf5283fa5ed]
```

😄

PR #51, opened 2019-07-18
Remove dependencies

devmotion:
This PR removes the dependencies on Cassette and SpecialFunctions. I'm currently running a test on Win32 (https://ci.appveyor.com/project/devmotion/delaydiffeq-jl-ungfr/build/job/femutot54m17paay) to check if by removing the dependency on Cassette we can get rid of the [compilation time issues of DelayDiffEq](https://github.com/JuliaDiffEq/DelayDiffEq.jl/issues/125).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -13,7 +13,6 @@ LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 VertexSafeGraphs = "19fa3120-7c27-5ec5-8db8-b0b0aa330d6f"
-Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
 
 [compat]
 julia = "1"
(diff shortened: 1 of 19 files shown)


devmotion:
It's still running, so unfortunately just removing Cassette does not seem to fix the issues in DelayDiffEq.

devmotion:
I merged the fix for the [Int32/Int64 issues](https://github.com/JuliaDiffEq/SparseDiffTools.jl/issues/53). The tests show that even with the reduced dependencies, the first test takes around 2-3 minutes to run on [Win64](https://ci.appveyor.com/project/devmotion/sparsedifftools-jl/build/job/n7sa4lgdeapw6eo4), Mac, and Linux 64bit and [around 1700 on Win32](https://ci.appveyor.com/project/devmotion/sparsedifftools-jl/build/job/ud0ax1ykc3nj21po).Both Win32 and Win64 tests throw a `EXCEPTION_ACCESS_VIOLATION`.

devmotion:
The tests on Win32 and Win64 with the Julia nightlies do not throw any exceptions. Could it be some upstream issue in Zygote? Could that also explain the long compilation times? According to the README

> Zygote's runtime performance should generally be good, but compile times are not optimised, so calling gradient the first time can have noticeable lag. BenchmarkTools is recommended to avoid measuring JIT time.

But still, the times are just too large.

devmotion:
I made Zygote an optional dependence. The only test that is running so far, the one [on Win32 (Julia 1.1.1)](https://ci.appveyor.com/project/devmotion/sparsedifftools-jl/build/job/2g3emryneragdvcj) indicates that the issues are caused by Zygote and/or the optional four functions `numback_hesvec, numback_hesvec!, autoback_hesvec, autoback_hesvec!`. Now the first test finishes in 117 sec (1744 sec before) but it currently hangs at the last test - in this test we load Zygote (and Cassette) for the first time.

devmotion:
@ChrisRackauckas Is this PR appropriate or should I change anything else? The tests seem to pass on AppVeyor now as well, I tested it in my own AppVeyor project: https://ci.appveyor.com/project/devmotion/sparsedifftools-jl

ChrisRackauckas:
We just need to update the README too.

PR #65, opened 2019-09-22
add oop jac `forwarddiff_color_jacobian`

huanglangwen:
```julia
forwarddiff_color_jacobian(oop_f, x, sparsity = nothing) == ForwardDiff.jacobian(oop_f, x)
jacout::Array{T,2} = forwarddiff_color_jacobian(oop_f, x, colorvec = ... , sparsity = ... )
jacout::T = forwarddiff_color_jacobian(oop_f, x, colorvec = ... , sparsity = ... , jac_prototype :: T = jac)
```


test/test_ad.jl
```diff
@@ -39,6 +50,17 @@ forwarddiff_color_jacobian!(_J1, f, x, colorvec = repeat(1:3,10))
 @test fcalls == 1
 
 fcalls = 0
+_J1 = forwarddiff_color_jacobian(oopf, x, colorvec = repeat(1:3,10), sparsity = _J, jac_prototype = _J)
```
ChrisRackauckas: we should test with Tracker.TrackedArray and StaticArrays. I think these will currently fail.

huanglangwen: Yes, but I have no idea about writing a real one.

huanglangwen: I made one, but I would expect a really bad performance. @ChrisRackauckas

src/differentiation/compute_jacobian_ad.jl
```diff
@@ -41,48 +41,83 @@ function ForwardColorJacCache(f,x,_chunksize = nothing;
...
+    chunksize = jac_cache.chunksize
+    color_i = 1
+    maxcolor = maximum(colorvec)
 
-    padding_matrix = BitMatrix(undef, length(x), padding_size)
-    partials = hcat(partials, padding_matrix)
+    vecx = vec(x)
+    
+    ncols=length(x)
+    J = jac_prototype isa Nothing ? zeros(Float64,ncols,ncols) : zero(jac_prototype)
```
ChrisRackauckas:
```suggestion
    J = jac_prototype isa Nothing ? false .* x .* x' : zero(jac_prototype)
```

src/differentiation/compute_jacobian_ad.jl
```diff
@@ -41,48 +41,83 @@ function ForwardColorJacCache(f,x,_chunksize = nothing;
...
+        partial_i = p[i]
+        t = reshape(Dual{typeof(f)}.(vecx, partial_i),size(t))
+        fx = f(t)
+        if !(sparsity isa Nothing)
+            for j in 1:chunksize
+                dx = vec(partials.(fx, j))
+                coords = [(rows_index[i],cols_index[i]) for i in 1:length(cols_index) if colorvec[cols_index[i]] == color_i ]
+                rows_index_c = [coord[1] for coord in coords]
+                cols_index_c = [coord[2] for coord in coords]
+                spmtx = collect(sparse(rows_index_c,cols_index_c,ones(length(rows_index_c)),size(J)...))
```
ChrisRackauckas: collecting into a sparse matrix will allocate, so we definitely don't want to do that. Should be a non-allocating kernel for static matrices.

huanglangwen:
Is it good for merge? @ChrisRackauckas

README.md
```diff
@@ -28,6 +28,17 @@ function f(dx,x)
   dx[end] = x[end-1] - 2x[end]
   nothing
 end
+
+function g(x) # out-of-place
+  global fcalls += 1
+  dx = zero(x)
```
ChrisRackauckas: we should make this test be on static arrays instead, so it tests that no mutation is used by failing on mutation

src/differentiation/compute_jacobian_ad.jl
```diff
@@ -41,48 +41,89 @@ function ForwardColorJacCache(f,x,_chunksize = nothing;
...
+                Ji = [j==cols_index_c[i] ? dx[i] : false for i in 1:ncols, j in 1:ncols]
+                J = J + Ji
+                color_i += 1
+                (color_i > maxcolor) && return J
+            end
+        else
+            for j in 1:chunksize
+                col_index = (i-1)*chunksize + j
+                (col_index > maxcolor) && return J
+                J = J + partials.(vec(fx), j) .* [(i==col_index ? 1 : 0) for i in 1:ncols]'
```
ChrisRackauckas: there should be a way to remove the array allocation here.

huanglangwen: I'm thinking of the delta tensor but that would introduce yet another dependency.

ChrisRackauckas: The delta tensor? Is that a FillArrays.jl object?

ChrisRackauckas:
To clarify the purpose here, the OOP version is really for two cases:

1. Static arrays. This is for very small systems and shouldn't allocate at all. This should be the fastest version of the function possible, so hopefully we the differentiation code won't slow it down.
2. Some forms of automatic differentiation do best with OOP code, and when they do so they will be allocating so we "don't need to care" about speed in this case.

So we really just need to make sure that `SArray` works and that case is fast.

src/differentiation/compute_jacobian_ad.jl
huanglangwen: I thought there would be some lazy Kronecker delta tensors, but looks like there's no such thing. FillArrays.jl doesn't work, the slice of eye generates sparse vector and its product with a normal array would be sparseCSC which is worse.

huanglangwen: I can't find any to make a general way of removing this array allocation. If we focus at `StaticArrays`, maybe I can make some overloads for `+` .

ChrisRackauckas: It would be fine to have a special branch for StaticArrays. How does ForwardDiff do it? It has a special routine for Jacobians that doesn't allocate.

PR #66, opened 2019-10-06
Correction for nonsquare jacobian

huanglangwen:


src/differentiation/compute_jacobian_ad.jl
```diff
@@ -80,8 +88,8 @@ function forwarddiff_color_jacobian(f,x::AbstractArray{<:Number},jac_cache::Forw
 
     vecx = vec(x)
     
-    ncols=length(x)
-    J = jac_prototype isa Nothing ? false .* x .* x' : zero(jac_prototype)
+    J = jac_prototype isa Nothing ? (sparsity isa Nothing ? false .* dx .* x' : zeros(eltype(x),size(sparsity))) : zero(jac_prototype)
```
ChrisRackauckas: `zeros` gives an `Array`

huanglangwen: I think when jac_prototype is not specified, J should be an Array.

ChrisRackauckas: true, makes sense.

PR #68, opened 2019-10-09
Missing setup for main example

jlperla:
But @huanglangwen  and @ChrisRackauckas  The next lines of code with `DiffEqDiffTools` don't work for me and I am not sure how to fix them.


jlperla:
I think my PR can be merged, but the other issues (where `finite_difference_jacobian!` doesn't work) should be looked at.

huanglangwen:
I think that’s because the sparsity is missing.

ChrisRackauckas:
Thanks, I cleaned up the rest here. It should now just copy/paste work

PR #76, opened 2019-11-25
[DO NOT MERGE] Test fixing tests with logging

matbesancon:
This sounds stupid but seems to pass


matbesancon:
So this is now passing tests, @ChrisRackauckas I have no idea how the behavior can be so weird. Cassette issue?

ChrisRackauckas:
Seems to be a blow up in compile times due to method invalidations. One of the signature changes is causing more methods to be invalidated at using time to cause more precompilation to happen, triggering failure. I think we should make sure we don't just change tests around that.

ChrisRackauckas:
I'm merging this since the compile time regressions seem to just be outside of here? and so this fixes master.

PR #77, opened 2019-11-26
Fix support of Julia 1.0

devmotion:
Fix https://github.com/JuliaDiffEq/SparseDiffTools.jl/issues/70 to support Julia 1.0 (and fix https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/974).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -6,6 +6,7 @@ version = "0.10.3"
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
 ArrayInterface = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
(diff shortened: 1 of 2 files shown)


devmotion:
In principle, one should run some CI tests on Julia 1.0 if one claims to support to support it :stuck_out_tongue: Should I add it to Travis?

ChrisRackauckas:
@pkj-m we really need to fix that stochastic bug. I think it may be due to uninitialized undefs here: https://github.com/JuliaDiffEq/SparseDiffTools.jl/blob/master/src/coloring/acyclic_coloring.jl#L19-L20 . Maybe initalizing with zeros or something would be safer and throw a correct error.

ChrisRackauckas:
Yeah... we should probably add Travis 1.0 tests. It just compiles so slow.

PR #79, opened 2019-11-29
min travis bump

matbesancon:


ChrisRackauckas:
This PR showed the other problems seem to actually be exhibited on master now, so I merged the green one and this will bump to 1.3. If this is green we're at least good, and it seems indeed that Cassette might be invalidating more methods than before.

PR #94, opened 2020-02-06
re-order checks

matbesancon:
I still have to run some  benchmarks there, but the function required a bit of cleanup


matbesancon:
Old version:
```julia
julia> @time matrix2graph2(jacsparsity, false)
671.048112 seconds (4.25 M allocations: 154.679 MiB, 0.02% gc time)

julia> @time matrix2graph2(jacsparsity, false)
663.122243 seconds (4.23 M allocations: 153.784 MiB, 0.01% gc time)
```

New version:
```julia
julia> @time matrix2graph(jacsparsity, false)
633.793763 seconds (1.02 M allocations: 100.897 MiB, 0.01% gc time)

julia> @time matrix2graph(jacsparsity, false)
606.861388 seconds (719.88 k allocations: 85.712 MiB, 0.02% gc time)
```

PR #96, opened 2020-03-06
Small API addition to forwarddiff_color_jacobian: kword chunksize

mauro3:


ChrisRackauckas:
Seems like MacOS is just slower? I'll dig into that.

mauro3:
Cool! Can you tag v1.4.0 and the re-run the test for https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/1054?

PR #98, opened 2020-04-05
Fix type instability due to mapslices in colorvec partial generation

ChrisRackauckas:
```julia
using BenchmarkTools, SparseArrays, SparseDiffTools, ForwardDiff
function f!(H, x)
    for n in 1:length(H)
        if isodd(n)
            H[n] = 2*x[1]
        else
            H[n] = 3*x[2]
        end
    end
end
H = zeros(10000)
x = 5.0 * ones(200)
f!(H, x)

pattern = sparse(vcat(collect(1:2:10000), collect(2:2:10000)), vcat(repeat([1], 5000), repeat([2], 5000)), 1, length(H), length(x))
jac = Float64.(sparse(pattern))
colors = matrix_colors(jac)
@btime forwarddiff_color_jacobian!(jac, f!, x, colorvec = colors);
```

goes from `9.993 ms (136398 allocations: 4.46 MiB)` to `605.200 μs (696 allocations: 284.89 KiB)`

### DIFF
--- a/src/differentiation/compute_jacobian_ad.jl
+++ b/src/differentiation/compute_jacobian_ad.jl
@@ -37,12 +37,8 @@ function ForwardColorJacCache(f,x,_chunksize = nothing;
         fx = similar(t)
         _dx = similar(x)
     else
-        pi = first(p) #perform trim (length(dx)<length(x)) or padding (length(dx)>length(x)) to first(p)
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
Tested locally due to SparsityDetection bug and passes, so merging.

Issue #93, opened 2020-02-04
Coloring benchmark problem


ChrisRackauckas:
```julia
using SparseDiffTools, LinearAlgebra, SparseArrays
N = 256
Mx = Tridiagonal([1.0 for i in 1:N-1],[-2.0 for i in 1:N],[1.0 for i in 1:N-1])
My = copy(Mx)
Iy = SparseMatrixCSC(I,N,N)
Ix = SparseMatrixCSC(I,N,N)
fJ = ones(3,3)
Dz = [1 0 0
      0 0 0
      0 0 0]
jacsparsity = kron(Dz,Iy,sparse(Mx)) + kron(Dz,sparse(My),Ix) + kron(fJ,Iy,Ix)
@time colorvec = matrix_colors(jacsparsity)
```

matbesancon:
what are Mx and My here?

ChrisRackauckas:
Oh, I cut https://github.com/JuliaDiffEq/DiffEqBenchmarks.jl/issues/91 a little too aggressively.

```julia
Mx = Tridiagonal([1.0 for i in 1:N-1],[-2.0 for i in 1:N],[1.0 for i in 1:N-1])
My = copy(Mx)
```

matbesancon:
For the record, recorded 600 seconds initially

matbesancon:
It's the whole matrix to graph function which is slow, the rest can wait:
```julia
julia> @time matrix2graph(jacsparsity, false)
475.198843 seconds (719.90 k allocations: 100.197 MiB, 0.01% gc time)
```

matbesancon:
two potential speedups already: `dropzeros` should be the inplace version, the non-inplace currently used returns a value that is ignored

matbesancon:
since the indices are known, `@inbounds` should be used throughout the function

matbesancon:
#94 is a first take on that, I think there may be a way to improve this even more

PR #107, opened 2020-05-12
reduce time complexity of matrix2graph

ghislainb:
This uses an intermediate structure to arrange the indices of the columns of the matrix by rows and avoid the nested for-loops of complexity O(nnz^2).

Benchmark #93 on my local machine, old version :
```
julia> @time matrix2graph(jacsparsity, false)
447.022331 seconds (1.02 M allocations: 100.974 MiB, 0.01% gc time)
```
New version :
```
julia> @time matrix2graph(jacsparsity, false)
  0.158604 seconds (1.18 M allocations: 116.212 MiB, 24.89% gc time)
```

_[my first Julia PR ! :smiley: ]_

### DIFF
--- a/src/coloring/matrix2graph.jl
+++ b/src/coloring/matrix2graph.jl
@@ -1,4 +1,31 @@
 """
+        _cols_by_rows(rows_index,cols_index)
+
+Returns a vector of rows where each row contains
+a vector of its column indices.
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
fantastic!

PR #128, opened 2021-01-09
Add method to mirror ForwardDiff's syntax for writing Jacobian of an oop function to an allocated matrix

chenwilliam77:
In ForwardDiff.jl, there is the ability to define an out-of-place function `f(x)` and autodiff it using
```
ForwardDiff.jacobian!(jac, f, x)
```
where `jac` is allocated outside of `ForwardDiff.jacobian!`. I have a specific use case where it'd be nice to have such a functionality in SparseDiffTools.jl and have implemented it. The tests should pass.

You may notice that I've defined
```
function forwarddiff_color_jacobian(J::SparseMatrixCSC{<:Number},f,x::AbstractArray{<:Number},jac_cache::ForwardColorJacCache,jac_prototype=nothing)
```
in addition to
```
function forwarddiff_color_jacobian(J::AbstractArray{<:Number},f,x::AbstractArray{<:Number},jac_cache::ForwardColorJacCache,jac_prototype=nothing)
```
I'm not too familiar with StaticArrays.jl, but I was unable to get one of the StaticArrays tests to pass when I used
```
J .+= (size(Ji)!=size(J) ? reshape(Ji,size(J)) : Ji)
```
instead of
```
J = J + (size(Ji)!=size(J) ? reshape(Ji,size(J)) : Ji)
```
To ensure I can mirror the function from ForwardDiff, I need the first line of code to work, but it seems to cause an error for StaticArrays. 

Thus, to avoid changing the behavior of the package by too much, I just defined a version of `forwarddiff_color_jacobian` specifically when `J` is a sparse matrix, since it should be safe to do `J .+= ...` The more general version of `forwarddiff_color_jacobian` continues to use `J += ...` If you think that it'd be better to switch the two (i.e. write a `forwarddiff_color_jacobian(J::StaticArray{<: Number}, ...)` and using `J .+= ..` for the general version), then I'm happy to update the code.

### DIFF
--- a/src/differentiation/compute_jacobian_ad.jl
+++ b/src/differentiation/compute_jacobian_ad.jl
@@ -189,7 +189,11 @@ function forwarddiff_color_jacobian(J::SparseMatrixCSC{<:Number},f,x::AbstractAr
                 cols_index_c = cols_index[pick_inds]
                 Ji = sparse(rows_index_c, cols_index_c, dx[rows_index_c],nrows,ncols)
                 # J = J + Ji
-                J .+= Ji
+                if j == 1 && i == 1
+                    J .= Ji # overwrite pre-allocated matrix
+                else
+                    J .+= Ji
+                end
                 color_i += 1
                 (color_i > maxcolor) && return J
             end
@@ -199,7 +203,11 @@ function forwarddiff_color_jacobian(J::SparseMatrixCSC{<:Number},f,x::AbstractAr
                 (col_index > ncols) && return J
                 Ji = mapreduce(i -> i==col_index ? partials.(vec(fx), j) : adapt(parameterless_type(J),zeros(eltype(J),nrows)), hcat, 1:ncols)
                 # J = J + (size(Ji)!=size(J) ? reshape(Ji,size(J)) : Ji) #branch when size(dx) == (1,) => size(Ji) == (1,) while size(J) == (1,1)
-                J .+= (size(Ji)!=size(J) ? reshape(Ji,size(J)) : Ji) #branch when size(dx) == (1,) => size(Ji) == (1,) while size(J) == (1,1)
+                if j == 1 && i == 1
+                    J .= (size(Ji)!=size(J) ? reshape(Ji,size(J)) : Ji) # overwrite pre-allocated matrix
+                else
+                    J .+= (size(Ji)!=size(J) ? reshape(Ji,size(J)) : Ji) #branch when size(dx) == (1,) => size(Ji) == (1,) while size(J) == (1,1)
+                end
             end
         end
     end
--- a/src/differentiation/compute_jacobian_ad.jl
+++ b/src/differentiation/compute_jacobian_ad.jl
@@ -78,7 +78,30 @@ end
     forwarddiff_color_jacobian(f,x,ForwardColorJacCache(f,x,chunksize,dx=dx,colorvec=colorvec,sparsity=sparsity),jac_prototype)
 end
 
+@inline function forwarddiff_color_jacobian(J::AbstractArray{<:Number}, f,
+                x::AbstractArray{<:Number};
+                colorvec = 1:length(x),
+                sparsity = nothing,
+                jac_prototype = nothing,
+                chunksize = nothing) # Note no dx keyword b/c can infer Jacobian's size via J
+    if sparsity === nothing && jac_prototype === nothing || !ArrayInterface.ismutable(x)
+        cfg = chunksize === nothing ? ForwardDiff.JacobianConfig(f, x) : ForwardDiff.JacobianConfig(f, x, ForwardDiff.Chunk(getsize(chunksize)))
+        return ForwardDiff.jacobian(f, x, cfg)
+    end
+    forwarddiff_color_jacobian(J,f,x,ForwardColorJacCache(f,x,chunksize,dx=similar(x,size(J, 1)),colorvec=colorvec,sparsity=sparsity),jac_prototype)
+end
+
 function forwarddiff_color_jacobian(f,x::AbstractArray{<:Number},jac_cache::ForwardColorJacCache,jac_prototype=nothing)
+    dx = jac_cache.dx
+    vecx = vec(x)
+    sparsity = jac_cache.sparsity
+
+    J = jac_prototype isa Nothing ? (sparsity isa Nothing ? false .* vec(dx) .* vecx' : zeros(eltype(x),size(sparsity))) : zero(jac_prototype)
+
+    forwarddiff_color_jacobian(J, f, x, jac_cache, jac_prototype)
+end
+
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
It might be good to separate the StaticArray case anyways by using `ArrayInterface.ismutable(x)`. That way we can have a fully non-allocating version and a fully non-mutating version.

src/differentiation/compute_jacobian_ad.jl
```diff
@@ -78,7 +78,31 @@ end
...
+@inline function forwarddiff_color_jacobian(J::AbstractArray{<:Number}, f,
+                x::AbstractArray{<:Number};
+                colorvec = 1:length(x),
+                sparsity = nothing,
+                jac_prototype = nothing,
+                chunksize = nothing,
+                dx = similar(x, size(J, 1))) #if dx is nothing, we will estimate dx at the cost of a function call
+    if sparsity === nothing && jac_prototype === nothing || !ArrayInterface.ismutable(x)
+        cfg = chunksize === nothing ? ForwardDiff.JacobianConfig(f, x) : ForwardDiff.JacobianConfig(f, x, ForwardDiff.Chunk(getsize(chunksize)))
+        return ForwardDiff.jacobian(f, x, cfg)
```
ChrisRackauckas, line 90: In this case it's not mutating `J` so this is incorrect?

ChrisRackauckas:
If SparseMatrixCSC has its own branch, then https://github.com/JuliaDiff/SparseDiffTools.jl/pull/128/files#diff-7e39b900cec80b7d52f89704e277c0f56afc3c2197d5ba4fef1fa9c943bcb0a2R135-R136 can be removed.

test/test_ad.jl
```diff
@@ -115,6 +115,14 @@ _J1 = forwarddiff_color_jacobian(oopf, x, colorvec = repeat(1:3,10), sparsity =
 @test _J1 ≈ J
 @test fcalls == 1
 
+#oop with in-place Jacobian
+fcalls = 0
+_oop_jacout = sparse(1.01 .* J) # want to be nonzero to check that the pre-allocated matrix is overwritten properly
+forwarddiff_color_jacobian(_oop_jacout, oopf, x; colorvec = repeat(1:3,10), sparsity = _J, jac_prototype = _J)
```
ChrisRackauckas, line 121: add a test on this call which is sparse but not SparseMatrixCSC, like BandedMatrix.

ChrisRackauckas:
>To ensure I can mirror the function from ForwardDiff, I need the first line of code to work, but it seems to cause an error for StaticArrays.

StaticArrays cannot be mutated, so they need a fully non-mutating path. We should simply throw an error if `ArrayInterface.mutable(J) == false`, since that means `J[i] = ...` will error.

ChrisRackauckas:
>Thus, to avoid changing the behavior of the package by too much, I just defined a version of forwarddiff_color_jacobian specifically when J is a sparse matrix, since it should be safe to do J .+= ... The more general version of forwarddiff_color_jacobian continues to use J += ... If you think that it'd be better to switch the two (i.e. write a forwarddiff_color_jacobian(J::StaticArray{<: Number}, ...) and using J .+= .. for the general version), then I'm happy to update the code.

We should do it slightly the other way around. Your version is a bit faster, so we should make your version be the one that is normally used, and we should send anything with `ArrayInterface.mutable(J) == false` to use the old OOP version since that would be slower for example on a dense `J`.

src/differentiation/compute_jacobian_ad.jl
```diff
@@ -150,14 +160,19 @@ function forwarddiff_color_jacobian(J::AbstractArray{<:Number},f,x::AbstractArra
...
+                end
             end
         end
     end
     J
 end
 
-function forwarddiff_color_jacobian(J::SparseMatrixCSC{<:Number},f,x::AbstractArray{<:Number},jac_cache::ForwardColorJacCache,jac_prototype=nothing)
+# When J is immutable, this version of forwarddiff_color_jacobian will avoid mutating J
+function forwarddiff_color_jacobian_immutable(J::AbstractArray{<:Number},f,x::AbstractArray{<:Number},jac_cache::ForwardColorJacCache)
```
ChrisRackauckas, line 175: because this is creating a new array, it cannot use the `J` in the first argument. It needs to return.

src/differentiation/compute_jacobian_ad.jl
```diff
@@ -94,16 +94,18 @@ end
...
+    if jac_prototype isa Nothing ? ArrayInterface.ismutable(x) : ArrayInterface.ismutable(jac_prototype)
+        # Whenever J is mutable, we mutate it to avoid allocations
+        dx = jac_cache.dx
+        vecx = vec(x)
+        sparsity = jac_cache.sparsity
 
-    if ArrayInterface.ismutable(J) # Whenever J is mutable, we mutate it to avoid allocations
+        J = jac_prototype isa Nothing ? (sparsity isa Nothing ? false .* vec(dx) .* vecx' :
+                                         zeros(eltype(x),size(sparsity))) : zero(jac_prototype)
         forwarddiff_color_jacobian(J, f, x, jac_cache)
```
ChrisRackauckas, line 106: this branch needs to `return J` afterwards as well, otherwise it's relying on the last line of `forwarddiff_color_jacobian` to return `J` which isn't necessarily guaranteed from a mutating function.

test/test_ad.jl
```diff
@@ -124,9 +124,11 @@ forwarddiff_color_jacobian(_oop_jacout, oopf, x; colorvec = repeat(1:3,10), spar
...
-_oop_jacout = BandedMatrix(1.01 .* _oop_jacout) # check w/BandedMatrix instead of sparse
+# BandedMatrix
+_oop_jacout = BandedMatrix(-1 => diag(J, -1) .* 1.01, 0 => diag(J, 0) .* 1.01,
+                           1 => diag(J, 1) .* 1.01) # check w/BandedMatrix instead of sparse
 fcalls = 0
-forwarddiff_color_jacobian(_oop_jacout, oopf, x; colorvec = repeat(1:3,10), sparsity = _J, jac_prototype = _J)
+forwarddiff_color_jacobian(_oop_jacout, oopf, x; colorvec = repeat(1:3,10), sparsity = _J)
 @test _oop_jacout ≈ J
 @test isa(_oop_jacout, BandedMatrix)
 @test fcalls == 1
```
ChrisRackauckas, line 134: A tridiagonal matrix should have 3 f calls? Especially with that colorvec.

chenwilliam77: What about the test at line 114-117? I didn't write that test, but it is essentially the same test I am running (in fact, that's what I based my original test off of).

ChrisRackauckas: oh wait, I guess it is one f call because of partial chunking. My bad.

ChrisRackauckas:
great work, thanks!

PR #139, opened 2021-04-11
Test Julia v1.5

ChrisRackauckas:


ChrisRackauckas:
All tests passing on master if v1.5, and same weird issue is on v1.6 https://github.com/JuliaDiff/SparseDiffTools.jl/issues/140 so tagging anyways.

Issue #144, opened 2021-07-15
matrix_colors can overwrite input (remove stored zeros)


sjdaines:
The culprit would appear to be:

     function matrix2graph(sparse_matrix::SparseMatrixCSC{<:Number, Int}, partition_by_rows::Bool=true)

        dropzeros!(sparse_matrix)

Not sure what the correct fix is.   Perhaps two reasonable options ?
1. test for stored zeros, and error if present, or
2. treat stored zeros as structural nonzeros

However silently calling dropzeros! and modifying the input is surely wrong...

ChrisRackauckas:
Oh, that `dropzeros!` is just wrong. Thanks for finding it. We should remove that. Coloring should only be w.r.t. structural zeros.

ChrisRackauckas:
Thanks a lot. This is a good find.

PR #145, opened 2021-07-18
matrix_colors include sparse matrix stored zeros (structural nonzeros)

sjdaines:
Fix for https://github.com/JuliaDiff/SparseDiffTools.jl/issues/144


ChrisRackauckas:
Thanks!

Issue #138, opened 2021-04-11
Large (20x) runtime difference between value and Jacobian with simple sparsity pattern


moyner:
As mentioned in issue #136, I am observing a discrepancy in performance between the AD and non-AD runtime that is larger than expected. Here is a MWE with a simple sparsity pattern that isolates the behavior and performs some sanity checks:
```julia
using ForwardDiff
using SparsityDetection, SparseArrays
using SparseDiffTools
using BenchmarkTools
using Test

n = 1000*1000
# n = 1000
L = [i for i in 1:n]
R = [i % n + 1 for i in 1:n]

function g(out, in)
    n = length(in)
    for i in eachindex(in)
        l = i
        r = i % n + 1
        out[i] = in[l] - in[r]
    end
end
inp = rand(n)
out = similar(inp)
# Time case without any AD
println("Calling value version:")
@btime g(out, inp)

pattern = jacobian_sparsity(g, out, inp)
jacA = Float64.(sparse(pattern)) # Output from non-cached version
jacB = Float64.(sparse(pattern)) # Output from JacCache'd version

colors = matrix_colors(jacA)
# Verify that it works without cache
println("Calling sparse AD without cache")
@btime forwarddiff_color_jacobian!(jacA, g, inp, colorvec = colors,
                                           sparsity = pattern)
# Set up cache
cache = ForwardColorJacCache(g, inp, colorvec = colors, 
                                     sparsity = pattern);
# Call with cache
println("Calling sparse AD with cache")
@btime forwarddiff_color_jacobian!(jacB, g, inp, cache)
# Check that they have same pattern & values
@test nnz(jacA-jacB) == 0
# Basic sanity check
@test maximum(colors) == 2
```
The function is a cyclical version of `diff` and the coloring is just 1,2 alternating with the Jacobian as a general sparse matrix. Filling the Jacobian should ideally be somewhere around 2-4 x the value function call from two AD passes since each Dual contains two floats if we disregard cache effects and the cost of dealing with the sparse data structure. Output for n = 2, 1000 and 1,000,000:
```julia
n = 1000000
```
```
Calling value version:
  25.803 ns (0 allocations: 0 bytes)
Explored path: SparsityDetection.Path(Bool[], 1)
Calling sparse AD without cache
  8.733 μs (83 allocations: 4.38 KiB)
Calling sparse AD with cache
  93.172 ns (0 allocations: 0 bytes)
```
```julia
n = 1000
```
```
[...]

Issue #100, opened 2020-04-06
Performance of tall Jacobians


MaximilianJHuber:
Let H be a matrix, x a vector, and f! a matrix-valued function. To make this fit the interface, I allow H to be a vector and reshape H :
```
using SparseArrays, SparseDiffTools, BenchmarkTools, ProfileView
function f!(H, x)
    if length(size(H)) == 1
        H = reshape(H, 500000, 4)
    end
    M = size(H)[1]
    N = size(H)[2]

    for m in 1:M
        for n in 1:N
            if isodd(n)
                H[m, n] = 2*x[1]
            else
                H[m, n] = 3*x[2]
            end
        end
    end
end
H = zeros(1000000, 4)
x = 5.0 * ones(200)
@views f!(H[:], x) # same as f!(H, x)
```
Of the 200 variables only the first two are relevant, but even they collapse to one color.
```
pattern = sparse([1, 3, 2, 4], [1, 1, 2, 2], 1, size(H, 2), length(x))
jac = Float64.(sparse(pattern))
colors = matrix_colors(jac)
pattern = kron(pattern, ones(size(H, 1)))
jac = Float64.(sparse(pattern))
cache = ForwardColorJacCache(f!, x, dx=H, colorvec=colors ,sparsity=pattern);
@btime forwarddiff_color_jacobian!(jac, f!, x, cache);
```
108.402 ms (4 allocations: 160 bytes)

It is still much faster than full ForwardDiff:
```
full_jac = Matrix(jac);
full_jac .= 0;
@time ForwardDiff.jacobian!(full_jac, f!, H, x);
```
  2.270 s (6 allocations: 198.39 MiB)

but `@profview forwarddiff_color_jacobian!(jac, f!, x, cache);` shows that most time is spent in `FiniteDiff._colorediteration!`.

YingboMa:
Here are some thoughts on optimizing `FiniteDiff._colorediteration!`. Currently, our iteration order is to linearly traverse the compressed Jacobian and to find the appropriate column in the uncompressed Jacobian to copy to. This could bad for cache locality because when the matrix is large and sparse, we want to minimize the number of times that we traverse the large uncompressed Jacobian. So, I suggest to linearly traverse the uncompressed Jacobian and to find the appropriate in compressed Jacobian to copy from.

MaximilianJHuber:
But isn't traversing the compressed Jacobian slower than traversing the uncompressed?

[...]

PR #146, opened 2021-07-18
Add fast path for forwarddiff_color_jacobian! with sparse J

sjdaines:
Partial fix for https://github.com/JuliaDiff/SparseDiffTools.jl/issues/138 https://github.com/JuliaDiff/SparseDiffTools.jl/issues/100

Add a fast path for the case where J and sparsity are are both SparseMatrixCSC
and have the same sparsity pattern.


src/differentiation/compute_jacobian_ad.jl
```diff
@@ -316,3 +333,17 @@ function forwarddiff_color_jacobian!(J::AbstractMatrix{<:Number},
     end
     return J
 end
+
+# fast version of FiniteDiff._colorediteration! for the case where J and sparsity have the same sparsity pattern
+@inline function _colorediteration!(Jsparsity::SparseMatrixCSC,vfx,colorvec,color_i,ncols)
```
ChrisRackauckas, line 338: Shouldn't this just go to FiniteDiff.jl?

sjdaines:
Probably... (I admit I don't I understand how this should work and feel like I'm proliferating cut-and-paste cases instead of sane dispatches).

PR https://github.com/JuliaDiff/FiniteDiff.jl/pull/124

Now we have:

1. add a _use_sparseCSC_common_sparsity!(J, sparsity) test function to FiniteDiff.iteration_utils.
2. move the new _colorediteration!(Jsparsity::SparseMatrixCSC,vfx,colorvec,color_i,ncols) to FiniteDiff.iteration_utils
3. FiniteDiff.finite_difference_jacobian! uses the same optimisation (although doesn't look like anyone has hit this problem there?)

src/differentiation/compute_jacobian_ad.jl
sjdaines: PR JuliaDiff/FiniteDiff.jl#124

ChrisRackauckas:
Add a lower bound on 2.8.1

PR #158, opened 2021-10-20
Add vecjac operators

avik-pal:


ChrisRackauckas:
We have Zygote under `@requires` right now for https://github.com/JuliaDiff/SparseDiffTools.jl/blob/master/src/differentiation/jaches_products_zygote.jl . Looking at the dependency list on JuliaHub and the stability of Zygote, I think we might want to keep it under Requires.jl for now.

src/SparseDiffTools.jl
```diff
@@ -8,6 +8,7 @@ using LightGraphs: SimpleGraph
 using Requires
 using VertexSafeGraphs
 using Adapt
+using Zygote
```
ChrisRackauckas, line 11:
```suggestion
```

src/differentiation/vecjac_products_zygote.jl
```diff
@@ -0,0 +1,53 @@
...
+    ϵ = sqrt(eps(real(T))) * max(one(real(T)), abs(norm(x)))
+    du = similar(x)
+    for i = 1:length(x)
+        x[i] += ϵ
+        f0 = f(x)
+        x[i] -= ϵ
+        du[i] = (((f0 .- _f0) ./ ϵ)'*vv)[1]
+    end
+    return vec(du)
+end
```
ChrisRackauckas, line 53: These ones don't need Zygote?

PR #161, opened 2021-11-01
Use SciMLBase instead of DiffEqBase

avik-pal:


src/differentiation/vecjac_products.jl
```diff
@@ -8,7 +18,7 @@ function num_vecjac!(
     cache2 = similar(v);
     compute_f0 = true,
 )
-    if SciMLBase.numargs(f) != 2
+    if _numargs(f) != 2
         du .= num_jacvec(f, x, v)
```
ChrisRackauckas, line 22: I would think this one would be `num_jacvec(f, du, x, v)` and then there would be one on the other dispatch which allocates? We can discuss this later though, I'll just merge and tag for now since you're using it.

PR #166, opened 2021-12-06
Remove numargs usage

ChrisRackauckas:
This just doesn't match the JuliaDiff style which is that the function call is defined by the function. I.e. out of place calls use out of place functions and in-place calls use in-place functions. This mixed it up which shouldn't be allowed in the API.


ChrisRackauckas:
@avik-pal

PR #167, opened 2021-12-08
Remove inline and specialize the function evaluations

ChrisRackauckas:
I believe the reason for the inlining was because the functions weren't specializing, and so this will fix that. Improves this ODE example a ton:

```julia
function rober2(u,p,t)
  y₁,y₂,y₃ = u
  k₁,k₂,k₃ = p
  du1 = -k₁*y₁+k₃*y₂*y₃
  du2 =  k₁*y₁-k₂*y₂^2-k₃*y₂*y₃
  du3 =  k₂*y₂^2
  SA[du1,du2,du3]
end
prob2 = ODEProblem{false}(rober2,SA[1.0,0.0,0.0],(0.0,1e5),SA[0.04,3e7,1e4])
@btime sol = solve(prob2,Rosenbrock23(),save_everystep=false)

```

### DIFF
--- a/src/differentiation/compute_jacobian_ad.jl
+++ b/src/differentiation/compute_jacobian_ad.jl
@@ -79,13 +79,13 @@ function generate_chunked_partials(x,colorvec,::Val{chunksize}) where chunksize
     chunked_partials
 end
 
-@inline function forwarddiff_color_jacobian(f,
+function forwarddiff_color_jacobian(f::F,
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
From `42.100 μs (506 allocations: 35.61 KiB)` to `14.900 μs (26 allocations: 3.28 KiB)`

PR #170, opened 2021-12-12
Allow users to set the tag for the configs

ChrisRackauckas:


ChrisRackauckas:
@chriselrod

src/differentiation/compute_jacobian_ad.jl
```diff
@@ -25,15 +27,21 @@ function ForwardColorJacCache(f::F,x,_chunksize = nothing;
         chunksize = _chunksize
     end
 
+    if tag === nothing
+        T = typeof(ForwardDiff.Tag(f,eltype(vec(x))))
```
chriselrod, line 31: Why not set it to a custom tag here?
DiffEq is already using a few custom tags.

ChrisRackauckas: This is for upstream packages to set package-specific tags.

PR #189, opened 2022-06-12
redesign automatic differentiation operators for SciMLOeprators

ChrisRackauckas:


YingboMa (commented):
`Base.typesof(a, b, c)` could replace `typeof.((a, b, c))`

Issue #142, opened 2021-05-29
Hessian decompression implementation


ElOceanografo:
Continuing on from this Discourse [discussion](https://discourse.julialang.org/t/laplace-approximation-for-mixed-models-in-julia-tmb-like-functionality/61796/14), I've implemented a method for decompressing a Hessian matrix from its coloring:
```julia
using SparsityDetection
using SparseDiffTools
using ForwardDiff
using LinearAlgebra
using SparseArrays

function sparsehess(f, g, hs, col, x)
    D = hcat([float.(i .== col) for i in 1:maximum(col)]...)
    δ = 1e-8
    ii, jj, vv = findnz(triu(hs))
    Hc = hcat([(g(x + δ * D[:, j]) - g(x)) / δ for j in 1:maximum(col)]...)
    H = sparse(ii, jj, zeros(length(vv)))
    for (i, j) in zip(ii, jj)
        H[i, j] = Hc[i, col[j]]
    end
    return Symmetric(H)
end
```
This can be polished, but I've got a couple of questions before working it up as a PR:
* Where/how should this go in the repo? Is there analogous Jacobian code I should model it off of?
* What methods for calculating the gradients should there be?
* On discourse @ChrisRackauckas suggested using star-2 coloring, but in some testing I just did it seems like the default `GreedyD1Color` is usually faster and more compact on Hessians with random sparsity patterns. Thoughts?
```julia
ntest = 500
nnzs = zeros(ntest)
coloring_timings = zeros(ntest, 2)
decompression_timings = zeros(ntest, 2)
ncolors = zeros(ntest, 2)
max_errors = zeros(ntest, 2)

colormethods = [SparseDiffTools.GreedyD1Color(), SparseDiffTools.GreedyStar2Color()]
for i in 1:ntest, 
    nx = 100
    xis = rand(1:nx, nx)
    xjs = rand(1:nx, nx)
    f(x) =  dot(x, x) + dot(x[xis], x[xjs])
    g(x) = ForwardDiff.gradient(f, x)
    x = randn(nx)
    hs = hessian_sparsity(f, x)
    nnzs[i] = nnz(hs)
    Hfd = ForwardDiff.hessian(f, x)
    for j in 1:2
        coloring_timings[i, j] = @elapsed col = matrix_colors(hs, colormethods[j])
        ncolors[i, j] = maximum(col)
        sparsehess(f, g, hs, col, x)
        t1 = time_ns()
[...]

PR #190, opened 2022-06-15
Add sparse Hessian decompression

ElOceanografo:
Following up on https://github.com/JuliaDiff/SparseDiffTools.jl/issues/142, this is a basic implementation of sparse Hessian decompression using matrix coloring. (See also https://github.com/SciML/Optimization.jl/issues/269.) Basic interface is modeled on the existing one for Jacobians; it includes `forwarddiff_color_hessian` and `forwarddiff_color_hessian!` methods for in-place and out-of-place computations, and a `ForwardColorHesCache` type containing buffers for computation. Feedback welcome!

To do:
- [x] Make sure there are method signatures corresponding to all the ones for Jacobians
- [x] More tests
- [x] Make sure performance is ok


Project.toml
```diff
@@ -15,6 +15,7 @@ Graphs = "86223c79-3864-5bf0-83f7-82e725a168b6"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
+SparsityDetection = "684fba80-ace3-11e9-3d08-3bc7ed6f96df"
```
ChrisRackauckas, line 18: Use Symbolics for this now.

src/differentiation/compute_hessian_ad.jl
```diff
@@ -0,0 +1,74 @@
...
+
+function forwarddiff_color_hessian!(H, 
+                                    f, 
+                                    g!, 
+                                    x, 
+                                    hes_cache::ForwardColorHesCache, 
+                                    d=sqrt(eps(Float64)))
+    for j in 1:hes_cache.ncolors
+        g!(hes_cache.G, x)
+        g!(hes_cache.dG, x + hes_cache.dx * @view hes_cache.D[:, j])
```
ChrisRackauckas, line 35: can we make this non-allocating?

Project.toml
ElOceanografo: I tried, following the examples in Symbolics, but it's not working as I'd expect:
```julia
julia> using Symbolics, LinearAlgebra

julia> f(x) = -dot(x, x) + 2 * x[2] * x[3]^2;

julia> @variables u[1:5];

julia> ff = f(u)
2(u[3]^2)*u[2] - (u[1]^2) - (u[2]^2) - (u[3]^2) - (u[4]^2) - (u[5]^2)

julia> Symbolics.hessian_sparsity(ff, u)
ERROR: TypeError: non-boolean (Num) used in boolean context
[...]
```
Also tried this, which doesn't seem to work either:
```julia
julia> Symbolics.hessian(ff, u) # all zeros, wrong
5×5 Matrix{Num}:
 0  0  0  0  0
 0  0  0  0  0
 0  0  0  0  0
 0  0  0  0  0
 0  0  0  0  0

julia> ForwardDiff.hessian(f, ones(5)) # correct
5×5 Matrix{Float64}:
 -2.0   0.0  0.0   0.0   0.0
  0.0  -2.0  4.0   0.0   0.0
  0.0   4.0  2.0   0.0   0.0
  0.0   0.0  0.0  -2.0   0.0
  0.0   0.0  0.0   0.0  -2.0
```
Am I just using Symbolics incorrectly?

ChrisRackauckas: Use `collect(u)` first?

ElOceanografo: 👍 Thanks, got it!

However, I now see SparsityDetection is already a dependency for some of the existing tests, which is also all I'm using it for (I added it to the main [deps] section of Project.toml by mistake). Unless you feel strongly about switching to Symbolics here, I think it makes more sense to just use the current dependency on SparsityDetection.

ElOceanografo:
`forwarddiff_color_jacobian!` has default arguments for `colorvec` and `sparsity`, which I currently don't supply for `forwarddiff_color_hessian!`. Currently not sure what makes the most sense: 
1) defaulting to an assumption that the Hessian is dense (i.e., `colorvec = eachindex(x), sparsity = ones(length(x), length(x))`), 
2) doing automatic sparsity detection when they aren't specified, or 
3) leaving it as-is and making the user explicitly supply the coloring and sparsity pattern.

ChrisRackauckas:
1

ElOceanografo:
I currently have two signatures for `forwarddiff_color_hessian!`:

* `forwarddiff_color_hessian!(H, f, x, colorvec, sparsity)` (creates a `ForwardColorHesCache`)
* `forwarddiff_color_hessian!(H, f, x, hes_cache)` (totally non-allocating)

There are also two corresponding signatures for `forward_color_hessian`, which create the sparse Hessian `H` instead of modifying it in place. I think this matches the interface for `forwarddiff_color_jacobian!`. Are any others needed?

ElOceanografo:
Ok, I think this is getting pretty close. @ChrisRackauckas when you've got a second, lemme know what else you think it needs and I'll try to take care of it!

README.md
```diff
@@ -207,6 +207,35 @@ function, otherwise it will become a dense matrix. If `jac_prototype` and
...
+fowrwarddif_color_hessian!(H, f, x, hescache)
+```
+
+By default, these methods use a mix of numerical and automatic differentiation,
+namely by taking finite differences of gradients calculated via ForwardDiff.jl.
+Alternatively, if you have your own custom gradient function `g!`, you can specify 
+it as an argument to `ForwardColorHesCache`:
+
+```julia
+hescache = ForwardColorHesCache(fscalar, x, g!, colorvec, sparsity)
```
ChrisRackauckas, line 233: optional argument last

src/differentiation/compute_hessian_ad.jl
```diff
@@ -0,0 +1,105 @@
...
+end
+
+
+function ForwardColorHesCache(f, 
+                              x::AbstractVector{<:Number},
+                              colorvec::AbstractVector{<:Integer}=eachindex(x),
+                              sparsity::Union{AbstractMatrix, Nothing}=nothing)
+    g!(G, x, grad_config) = ForwardDiff.gradient!(G, f, x, grad_config)
+    return ForwardColorHesCache(f, x, g!, colorvec, sparsity)
+end
```
ChrisRackauckas, line 57: collapse this into above by making `g` an optional argument with that as its argument

src/differentiation/compute_hessian_ad.jl
```diff
@@ -0,0 +1,105 @@
...
+    for j in 1:hes_cache.ncolors
+        hes_cache.grad!(hes_cache.G, x, hes_cache.grad_config)
+        x .+= ϵ .* @view hes_cache.D[:, j]
+        hes_cache.grad!(hes_cache.dG, x, hes_cache.grad_config)
+        x .-= ϵ .* @view hes_cache.D[:, j]
+        hes_cache.buffer[:, j] .= (hes_cache.dG .- hes_cache.G) ./ ϵ
+    end
+    ii, jj, vv = findnz(hes_cache.sparsity)
+    for (i, j) in zip(ii, jj)
+        H[i, j] = hes_cache.buffer[i, hes_cache.colors[j]]
```
ChrisRackauckas, line 74: should probably `fill!(H,false)` for safety, and have a kwarg for the safety switch?

src/differentiation/compute_hessian_ad.jl
```diff
@@ -0,0 +1,97 @@
...
+                                    x::AbstractArray{<:Number}, 
+                                    hes_cache::ForwardColorHesCache;
+                                    safe = true)
+    ϵ = cbrt(eps(eltype(x)))
+    for j in 1:hes_cache.ncolors
+        hes_cache.grad!(hes_cache.G, x, hes_cache.grad_config)
+        x .+= ϵ .* @view hes_cache.D[:, j]
+        hes_cache.grad!(hes_cache.dG, x, hes_cache.grad_config)
+        x .-= ϵ .* @view hes_cache.D[:, j]
+        hes_cache.buffer[:, j] .= (hes_cache.dG .- hes_cache.G) ./ ϵ
```
ChrisRackauckas, line 59: From this naming I would've expected ForwardDiff here.  Shouldn't this be `num_over_forward_color_hessian!`, and we should have a `forward_over_forward_color_hessian!`?

ElOceanografo: How about `num_forward_color_hessian!` or `numauto_color_hessian!` so it's similar to the `*_hesvec` methods?

ChrisRackauckas: `numauto_color_hessian!` is good

src/differentiation/compute_hessian_ad.jl
```diff
@@ -0,0 +1,97 @@
...
```
ChrisRackauckas, line 59:
```suggestion
        hes_cache.grad!(hes_cache.dG, x, hes_cache.grad_config)
        x .-= 2ϵ .* @view hes_cache.D[:, j]
        hes_cache.buffer[:, j] .= (hes_cache.dG .- hes_cache.G) ./ 2ϵ
```

If you're not going to use a cached `f` here, is there a reason to not do a central difference?

ChrisRackauckas:
Is this still WIP? Looks close to done.

src/differentiation/compute_hessian_ad.jl
ElOceanografo: The only reason not to is because it modifies `x` three times as opposed to two, saving a little time. However, profiling shows that the time saved is only ~2% of the total, and the centered difference decreases error by more than 10x, so a worthwhile tradeoff. Changed as suggested.

Can you clarify what you mean by "use a cached `f`?"

PR #205, opened 2022-10-09
remove deprecated SparsityDetection

ArnoStrouwen:
Removes reliance on deprecated `SparsityDetection.jl` package.

A decision has to be made if we want to directly support sparsity detection here, through a small wrapper around the newer functionality in  `Symbolics.jl`.


src/differentiation/sparsity_detection.jl
```diff
@@ -0,0 +1,8 @@
+function jacobian_sparsity(func::Function, output::AbstractVector{T}, input::AbstractVector{T}; kwargs...) where {T<:Number}
+    Symbolics.jacobian_sparsity(func,output,input)
+end
+function hessian_sparsity(func::Function, input::AbstractVector{T}; kwargs...) where {T<:Number}
+    vars = map(Symbolics.variable, eachindex(input))
+    expr = func(vars; kwargs...)
+    Symbolics.hessian_sparsity(expr, vars)
+end
```
ChrisRackauckas, line 8: These should go to Symbolics.jl. It's type piracy here.

ChrisRackauckas:
Add the extra helper functions to Symbolics.jl, with some docs. It should really only be tested here.

ChrisRackauckas:
Released, so try removing now

ChrisRackauckas (approved):

PR #208, opened 2022-10-25
SciML style formatting

ArnoStrouwen:


src/coloring/acyclic_coloring.jl
```diff
@@ -187,10 +204,12 @@ end
...
-function find(w::Integer,
-              x::Integer,
-              g::Graphs.AbstractGraph,
-              two_colored_forest::DisjointSets{<:Integer})
+function find(
+    w::Integer,
+    x::Integer,
+    g::Graphs.AbstractGraph,
+    two_colored_forest::DisjointSets{<:Integer},
+)
```
ChrisRackauckas, line 212: This isn't SciMLStyle. Is the formatter toml missing?

src/coloring/acyclic_coloring.jl
```diff
@@ -30,7 +30,14 @@ function color_graph(g::Graphs.AbstractGraph, ::AcyclicColoring)
...
                         if forbidden_colors[color[x]] != v
-                            prevent_cycle!(first_visit_to_tree,forbidden_colors,v, w, x, g, two_colored_forest,color)
+                            prevent_cycle!(first_visit_to_tree,
+                                           forbidden_colors,
+                                           v,
+                                           w,
+                                           x,
+                                           g,
+                                           two_colored_forest,
+                                           color)
```
ChrisRackauckas, line 40: did you start from scratch after adding the toml?

ArnoStrouwen: What does scratch mean? I added the toml and called format on the directory again.
This changed most files and I made another commit.

ChrisRackauckas: yeah you cannot do that because running it with a different style and then this style does not give the same result as running it only with SciML style.

ArnoStrouwen: The explanation here gave me the idea that it would be exactly the same end result no matter the starting point.
https://domluna.github.io/JuliaFormatter.jl/stable/how_it_works/#Generating-an-FST

ChrisRackauckas: Yeah but the formatting isn't commutative, so it actually doesn't work out like that if you have two different styles 😅

PR #212, opened 2023-02-04
make AD operators scimloperators

vpuri3:
Plan: https://github.com/SciML/SciMLOperators.jl/issues/142


vpuri3:
@ChrisRackauckas @YingboMa can you take a look? Maybe suggest some tests for `VecJacProd` operator

vpuri3:
TODO - add 5 arg `mul!` in test, and add tests for `VecJacProd`.

vpuri3:
TODO - write 5-arg mul! tests after https://github.com/SciML/SciMLOperators.jl/pull/150

vpuri3:
@ChrisRackauckas this should be good. Please take a look

test/runtests.jl
```diff
@@ -12,16 +12,18 @@ function activate_gpu_env()
...
+    #@time @safetestset "Exact coloring via contraction" begin include("test_contraction.jl") end
+    #@time @safetestset "Greedy distance-1 coloring" begin include("test_greedy_d1.jl") end
+    #@time @safetestset "Greedy star coloring" begin include("test_greedy_star.jl") end
+    #@time @safetestset "Acyclic coloring" begin include("test_acyclic.jl") end
+    #@time @safetestset "Matrix to graph conversion" begin include("test_matrix2graph.jl") end
+    #@time @safetestset "AD using colorvec vector" begin include("test_ad.jl") end
+    #@time @safetestset "Hessian colorvecs" begin include("test_sparse_hessian.jl") end
+    #@time @safetestset "Integration test" begin include("test_integration.jl") end
+    #@time @safetestset "Special matrices" begin include("test_specialmatrices.jl") end
+    #@time @safetestset "Jac Vecs and Hes Vecs" begin include("test_jaches_products.jl") end
```
ChrisRackauckas, line 24: ???????

vpuri3: sorry, fixed it. :)

vpuri3:
@ChrisRackauckas

ChrisRackauckas:
That looks fine in general, but why is it not just replacing the old ones? Delete the old ones and use their name?

vpuri3:
OrdinaryDiffEq uses `JecVac` in `src/derivative_utils.jl`. So deleting the old ones would break stuff there.

ChrisRackauckas:
This should be a drop-in replacement though?

vpuri3:
No. ODE.jl uses the type `JacVec`. Eg
```julia
if W.J !== nothing && !(W.J isa SparseDiffTools.JacVec) 
```
I don't have the type JacVec anymore. My functions `JacVecProd, etc` dispatch to `FunctionOperator`

ChrisRackauckas:
Then we can make this a breaking release v2, with this change and adding the ArrayInterface packages from the README as extension packages.

src/differentiation/jaches_products.jl
```diff
@@ -198,112 +198,105 @@ end
 
 ### Operator Forms
 
-struct JacVec{F, T1, T2, xType}
+mutable struct FwdModeAutoDiffVecProd{F,U,C,V,V!} <: AbstractAutoDiffVecProd
```
ChrisRackauckas, line 201: use the original names

vpuri3: The diff isn't conveying things correctly - you might want to look at the files. `FwdModeAutoDiffVecProd, RevModeAutoDiffVecProd` are not user facing structs. They are not exported.

What are exported are the functions `JacVec, HesVec, HesVecGrad, VecJac`. These are the original names.

eg JacVec is
https://github.com/JuliaDiff/SparseDiffTools.jl/blob/4d7ce81bf3663e742dadfd560b6697b9edb7052f/src/differentiation/jaches_products.jl#L226-L250

vpuri3: @ChrisRackauckas

Project.toml
```diff
@@ -14,8 +14,10 @@ ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
 Graphs = "86223c79-3864-5bf0-83f7-82e725a168b6"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Requires = "ae029012-a4dd-5104-9daa-d747884805df"
+SciMLOperators = "c0aeaf25-5076-4817-a8d5-81caf7dfa961"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
+Tricks = "410a4b4d-49e4-4fbc-ab6d-cb71b17b3775"
```
ChrisRackauckas, line 20: missing a compat

test/test_jaches_products.jl
```diff
@@ -65,61 +65,90 @@ cache4 = ForwardDiff.Dual{typeof(ForwardDiff.Tag(Nothing, eltype(x))), eltype(x)
...
+    prob = ODEProblem(ff, u0, tspan)
+    @test solve(prob, TRBDF2()).retcode == :Success
+    @test solve(prob, TRBDF2(linsolve = KrylovJL_GMRES())).retcode == :Success
+    @test solve(prob, Exprb32()).retcode == :Success
+    @test solve(prob, Rosenbrock23()).retcode == :Success
+    @test solve(prob, Rosenbrock23(linsolve = KrylovJL_GMRES())).retcode == :Success
+end
+=#
+
+# HesVec
```
ChrisRackauckas, line 105: ?

vpuri3: OrdinaryDiffEq fails to precompile with this version of SparseDiffTools because of the changes we have discussed above. So I'm going to remove these tests and put them back in once we make the requisite changes in OrdianryDiffEq (in `OrdinaryDiffEq.jl/src/derivative_utils.jl`). Making a note of that here: https://github.com/SciML/SciMLOperators.jl/issues/142

src/differentiation/jaches_products.jl
```diff
@@ -198,112 +198,105 @@ end
...
     else
-        num_hesvecgrad!(dy, L.g, L.x, v, L.cache1, L.cache2)
+        cache1 = similar(u)
+        cache2 = similar(u)
     end
+
+    cache = (cache1, cache2,)
+
+    vecprod  = autodiff ? auto_hesvecgrad  : num_hesvecgrad
+    vecprod! = autodiff ? auto_hesvecgrad! : num_hesvecgrad!
```
ChrisRackauckas, line 293: `f` will not necessarily support both. They have different calling conventions. The function needs to be checked before declaring this.

src/differentiation/vecjac_products.jl
```diff
@@ -34,3 +34,80 @@ function num_vecjac(f, x, v, f0 = nothing)
...
+
+    isinplace  = static_hasmethod(f, typeof((u, p, t)))
+    outofplace = static_hasmethod(f, typeof((u, u, p, t)))
+
+    if !(isinplace) & !(outofplace)
+        error("$f must have signature f(u, p, t), or f(du, u, p, t)")
+    end
+
+    L = RevModeAutoDiffVecProd(f, u, cache, vecprod, vecprod!; autodiff = autodiff,
+                               isinplace = isinplace, outofplace = outofplace)
```
ChrisRackauckas, line 105: The other function operators need this treatment too

src/differentiation/vecjac_products.jl
```diff
@@ -34,3 +34,80 @@ function num_vecjac(f, x, v, f0 = nothing)
...
+    L.vecprod!(dv, (_du, _u) -> L.f(_du, _u, p, t), L.u, v, L.cache...)
+end
+
+function VecJac(f, u::AbstractArray, p = nothing, t = nothing; autodiff = true,
+                ishermitian = false, opnrom = true)
+
+    cache = (similar(u), similar(u),)
+
+    vecprod  = autodiff ? auto_vecjac  : num_vecjac
+    vecprod! = autodiff ? auto_vecjac! : num_vecjac!
```
ChrisRackauckas, line 95: This should also have a choice for Zygote. For stability, JvFormZygote() and such? Use the same one the Jv operators.

vpuri3:
currently, `JacVec, HesVec, HesVecGrad` do not use `p`, `t`. They accept functions like `f(x), f(y, x)`.

```julia
function (L::FwdModeAutoDiffVecProd)(v, p, t)
    L.vecprod(L.f, L.u, v)
end
```

As compared to `VecJac` that accepts `f(u, p, t), f(du, u, p, t)`
```julia
# Interpret the call as df/du' * u
function (L::RevModeAutoDiffVecProd)(v, p, t)
    L.vecprod(_u -> L.f(_u, p, t), L.u, v)
end
```

Do we want to enforce the same `(u,p,t)` signature in the latter?

ChrisRackauckas:
No, these don't have p and t.

vpuri3:
ok, then this is complete. needs to rerun ci once array interface stuff is cleared out

src/differentiation/jaches_products.jl
```diff
@@ -198,112 +198,126 @@ end
 
 ### Operator Forms
 
-struct JacVec{F, T1, T2, xType}
+mutable struct FwdModeAutoDiffVecProd{F,U,C,V,V!} <: AbstractAutoDiffVecProd
```
ChrisRackauckas, line 201: why is this a mutable struct?

src/differentiation/jaches_products.jl
```diff
@@ -198,112 +198,126 @@ end
...
+
+    vecprod  = autodiff ? numauto_hesvec  : num_hesvec
+    vecprod! = autodiff ? numauto_hesvec! : num_hesvec!
+
+    outofplace = static_hasmethod(f, typeof((u,)))
+    isinplace  = static_hasmethod(f, typeof((u,)))
+
+    if !(isinplace) & !(outofplace)
+        error("$f must have signature f(u).")
+    end
```
ChrisRackauckas, line 281: there is only out of place here since it returns a scalar?

ChrisRackauckas: oh I guess as an operator it is in-place as well, I see.

src/differentiation/vecjac_products.jl
ChrisRackauckas: I'll do this as a follow up, it should be included in the breaking update.

src/differentiation/vecjac_products_zygote.jl
```diff
@@ -7,3 +7,6 @@ function auto_vecjac(f, x, v)
     vv, back = Zygote.pullback(f, x)
     return vec(back(reshape(v, size(vv)))[1])
 end
+
+const ZygoteVecJac = VecJac
```
ChrisRackauckas, line 11: what's this for?

ChrisRackauckas:
Just two small questions are left. After this merges, then we should change the autodiff choices to be non-boolean logic and Zygote an extension package in order to release.

vpuri3:
@ChrisRackauckas about the zygote stuff...

You mentioned above that we need Zygote operators for stability. I wrote functions `ZygoteVecJac` and `ZygoteHesVec` that use `num/auto_vecjac(!)`, `num/autoback_hesvec(!)`.

The difference between `VecJac` and `ZygoteVecJac` is that the former defaults to `autodiff = false` and the later `true`. That way `VecJac` won't error with default kwargs if zygote isn't loaded. Since `VecJac` uses `auto_vecjac(!)` when `autodiff = true`, I included a warning in there to make sure Zygote is loaded:

```julia
    if autodiff 
        @assert isdefined(SparseDiffTools, :auto_vecjac) "Please load Zygote with `using Zygote`, or `import Zygote` to use VecJac with `autodiff = true`."
    end
```

ChrisRackauckas:
> You mentioned above that we need Zygote operators for stability. I wrote functions ZygoteVecJac and ZygoteHesVec that use num/auto_vecjac(!), num/autoback_hesvec(!).

No, I meant that `autodiff` shouldn't be boolean logic but multi-valued logic through types choices, like `AutoZygote()`.

ChrisRackauckas:
We can follow up with the extension package stuff.

PR #214, opened 2023-02-18
rm ArrayInterfaceStaticArrays

vpuri3:
@ChrisRackauckas do we need to add `StaticArrayInterfaces`?


vpuri3:
```julia
(SparseDiffTools) pkg> add StaticArrayInterface
    Updating registry at `~/.julia/registries/General.toml`
ERROR: The following package names could not be resolved:
 * StaticArrayInterface (not found in project, manifest or registry)
```

Issue #215, opened 2023-02-19
Todos for v2


ChrisRackauckas:
- [ ] Make Zygote an extension package
- [ ] Make autodiff a multi-valued logic 

@Vaibhavdixit02 I think we may want to move `AutoZyogte()` and etc. types to a package that both this and Optimization.jl build off of?

Vaibhavdixit02:
Yes, that will be quite useful for Optimization.jl

vpuri3:
Hi @Vaibhavdixit02 @ChrisRackauckas, any news on this front? Thanks! :D

Vaibhavdixit02:
I'll create the repo today @vpuri3

Vaibhavdixit02:
https://github.com/JuliaRegistries/General/pull/78584

PR #221, opened 2023-03-12
multivalue autodiff

vpuri3:
close https://github.com/JuliaDiff/SparseDiffTools.jl/issues/215

### DIFF
--- a/ext/SparseDiffToolsZygote.jl
+++ b/ext/SparseDiffToolsZygote.jl
@@ -7,6 +7,7 @@ if isdefined(Base, :get_extension)
     using ForwardDiff: ForwardDiff, Dual, partials
     using SciMLOperators: FunctionOperator
     using Tricks: static_hasmethod
+    using ADTypes
 else
     import ..Zygote
     using ..LinearAlgebra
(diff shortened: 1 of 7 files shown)


vpuri3:
@ChrisRackauckas good to go

ext/SparseDiffToolsZygote.jl
```diff
@@ -115,8 +116,8 @@ function SparseDiffTools.auto_vecjac(f, x, v)
     return vec(back(reshape(v, size(vv)))[1])
 end
 
-function SparseDiffTools.ZygoteVecJac(args...; autodiff = true, kwargs...)
-    VecJac(args...; autodiff = autodiff, kwargs...)
+function SparseDiffTools.ZygoteVecJac(args...; kwargs...)
```
ChrisRackauckas, line 119: this should just be renamed to VecJac, and AutoZygote is the Zygote version of it (but then we want to make Enzyme etc. versions soon.

src/differentiation/jaches_products.jl
```diff
@@ -306,6 +280,8 @@ function BackHesVec(f, u::AbstractArray, p = nothing, t = nothing; autodiff = Au
         cache2 = copy(u)
 
         (cache1, cache2), autoback_hesvec, autoback_hesvec!
+    else
+        @error("Set autodiff to either AutoForwardDiff(), or AutoFiniteDiff()")
```
ChrisRackauckas, line 284:
```suggestion
        @error("Set autodiff to either AutoForwardDiff(), AutoZygote(), or AutoFiniteDiff()")
```

PR #222, opened 2023-03-12
Test ZygoteHesVec

vpuri3:
looks like I did a typo


vpuri3:
@ChrisRackauckas good to go

test/test_jaches_products.jl
```diff
@@ -127,7 +127,7 @@ update_coefficients!(L, v, nothing, 0.0)
 @test mul!(dy, L, v)≈numback_hesvec(g, v, v) rtol=1e-2
 dy=rand(N);_dy=copy(dy);@test mul!(dy,L,v,a,b) ≈ a*numback_hesvec(g,x,v) + b*_dy rtol=1e-2
 
-L = HesVec(g, x)
+L = ZygoteHesVec(g, x)
```
ChrisRackauckas, line 130: how was this typo not causing a failure before?

vpuri3: `HesVec` defaults to finitediff, so the tests just passed with whatever tolerance was set. I had copied the code block from a few lines above which had passed. So the result was that zygote versions of hesvecv products weren't tested.

PR #232, opened 2023-03-27
have update_coeffs(L::ADVecProd,) recursively update L.f

vpuri3:
https://github.com/SciML/OrdinaryDiffEq.jl/pull/1917

### DIFF
--- a/src/differentiation/jaches_products.jl
+++ b/src/differentiation/jaches_products.jl
@@ -207,10 +207,12 @@ struct FwdModeAutoDiffVecProd{F,U,C,V,V!} <: AbstractAutoDiffVecProd
 end
 
 function update_coefficients(L::FwdModeAutoDiffVecProd, u, p, t)
-    FwdModeAutoDiffVecProd(L.f, u, L.vecprod, L.vecprod!, L.cache)
+    f = update_coefficients(L.f, u, p, t)
+    FwdModeAutoDiffVecProd(f, u, L.vecprod, L.vecprod!, L.cache)
 end
 
 function update_coefficients!(L::FwdModeAutoDiffVecProd, u, p, t)
+    update_coefficients!(L.f, u, p, t)
     copy!(L.u, u)
     L
 end
--- a/src/differentiation/jaches_products.jl
+++ b/src/differentiation/jaches_products.jl
@@ -208,7 +208,7 @@ end
 
 function update_coefficients(L::FwdModeAutoDiffVecProd, u, p, t)
     f = update_coefficients(L.f, u, p, t)
-    FwdModeAutoDiffVecProd(f, u, L.vecprod, L.vecprod!, L.cache)
+    FwdModeAutoDiffVecProd(f, u, L.cache, L.vecprod, L.vecprod!)
 end
 
 function update_coefficients!(L::FwdModeAutoDiffVecProd, u, p, t)
--- a/src/differentiation/jaches_products.jl
+++ b/src/differentiation/jaches_products.jl
@@ -218,10 +218,12 @@ function update_coefficients!(L::FwdModeAutoDiffVecProd, u, p, t)
 end
 
 function (L::FwdModeAutoDiffVecProd)(v, p, t)
+    L = update_coefficients(L, v, p, t)
     L.vecprod(L.f, L.u, v)
 end
 
 function (L::FwdModeAutoDiffVecProd)(dv, v, p, t)
+    update_coefficients!(L, v, p, t)
     L.vecprod!(dv, L.f, L.u, v, L.cache...)
 end
 
--- a/src/differentiation/jaches_products.jl
+++ b/src/differentiation/jaches_products.jl
@@ -218,12 +218,10 @@ function update_coefficients!(L::FwdModeAutoDiffVecProd, u, p, t)
 end
 
 function (L::FwdModeAutoDiffVecProd)(v, p, t)
-    L = update_coefficients(L, v, p, t)
     L.vecprod(L.f, L.u, v)
 end
 
 function (L::FwdModeAutoDiffVecProd)(dv, v, p, t)
-    update_coefficients!(L, v, p, t)
     L.vecprod!(dv, L.f, L.u, v, L.cache...)
 end
 
--- a/src/differentiation/jaches_products.jl
+++ b/src/differentiation/jaches_products.jl
@@ -78,6 +78,7 @@ function num_hesvec!(dy,
     g(cache2, x)
     @. x -= 2ϵ * v
     g(cache3, x)
+    @. x += ϵ * v
     @. dy = (cache2 - cache3) / (2ϵ)
 end
 
@@ -110,6 +111,7 @@ function numauto_hesvec!(dy,
     g(cache1, x)
     @. x -= 2ϵ * v
     g(cache2, x)
+    @. x += ϵ * v
     @. dy = (cache1 - cache2) / (2ϵ)
 end
 
@@ -158,6 +160,7 @@ function num_hesvecgrad!(dy, g, x, v, cache2 = similar(v), cache3 = similar(v))
     g(cache2, x)
     @. x -= 2ϵ * v
     g(cache3, x)
+    @. x += ϵ * v
     @. dy = (cache2 - cache3) / (2ϵ)
 end
 
--- a/src/differentiation/vecjac_products.jl
+++ b/src/differentiation/vecjac_products.jl
@@ -77,16 +77,16 @@ end
 
 # Interpret the call as df/du' * u
 function (L::RevModeAutoDiffVecProd)(v, p, t)
-    L.vecprod(_u -> L.f(_u, p, t), L.u, v)
+    L.vecprod(L.f, L.u, v)
 end
 
 # prefer non in-place method
 function (L::RevModeAutoDiffVecProd{ad,iip,true})(dv, v, p, t) where{ad,iip}
-    L.vecprod!(dv, _u -> L.f(_u, p, t), L.u, v, L.cache...)
+    L.vecprod!(dv, L.f, L.u, v, L.cache...)
 end
 
 function (L::RevModeAutoDiffVecProd{ad,true,false})(dv, v, p, t) where{ad}
-    L.vecprod!(dv, (_du, _u) -> L.f(_du, _u, p, t), L.u, v, L.cache...)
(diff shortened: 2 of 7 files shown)


vpuri3:
@gaurav-arya, can you take this over and write a few tests in `test/test_jaches_products.jl` that utilize this functionality?

ie

change
```julia
f(y, x) = mul!(y, A, x)
f(x) = A * x
```
to something something like UJacobianWrapper
```julia
struct Func
  u
  p
  t
end

function update_coefficients!(f::Fun, u, p, t)
# update u, p, t
end

function update_coefficients(f::Fun, u, p, t)
# update u, p, t
end

(f::Fun)(x) = t * p * u
```

src/differentiation/jaches_products.jl
```diff
@@ -207,10 +207,12 @@ struct FwdModeAutoDiffVecProd{F,U,C,V,V!} <: AbstractAutoDiffVecProd
 end
 
 function update_coefficients(L::FwdModeAutoDiffVecProd, u, p, t)
-    FwdModeAutoDiffVecProd(L.f, u, L.vecprod, L.vecprod!, L.cache)
+    f = update_coefficients(L.f, u, p, t)
```
gaurav-arya, line 210: hmm, so you're actually going for solution quite similar to https://github.com/SciML/SciMLOperators.jl/issues/172, right? i.e. you're not changing the signature of `f`, that stays `f(u)`, right?

gaurav-arya: for this soln to work, we need to add this line in `FunctionOperator's update_coefficients, right:
```
# Allow L.op to have u/p/t dependence (safe since update_coefficients! defaults to null-op)
update_coefficients!(L.op, u, p, t)
```
Otherwise the update coefficients for `FwdModeAutoDiffVecProd` etc. won't get called

vpuri3: FwdModeAutoDiffVecProd is like a wrapper that manages the difference bw signatures. It's `mul` uses `auto_jacvec` that expects signature `f(u), f!(v, u)`

vpuri3: FunctionOperator's update_coeffs does that already. pasting below

```julia
function update_coefficients!(L::FunctionOperator, u, p, t)
    ops = getops(L)
    for op in ops
        update_coefficients!(op, u, p, t)
    end     

    L.p = p
    L.t = t

    nothing
end
```

gaurav-arya: ah, didn't realize that `op` was already part of `L`'s ops, great:)

vpuri3: same for oop `update_coeffs`

vpuri3:
thanks. can you also add these tests to `vecjac`?

gaurav-arya:
@vpuri3 is the CI working properly? It doesn't seem to have caught the bugs due to ArrayInterface v7, and moreover there's a ref to an object here that doesn't exist:

https://github.com/JuliaDiff/SparseDiffTools.jl/blob/f191085e7ab4c5281fb80e4c779ae440b5dcc3cc/test/test_vecjac_products.jl#L27

vpuri3:
remove zygotejacvec

vpuri3:
file an issue to fix ci after this pr please

gaurav-arya:
#234

test/test_jaches_products.jl
```diff
@@ -67,21 +79,21 @@ cache4 = ForwardDiff.Dual{typeof(ForwardDiff.Tag(Nothing, eltype(x))), eltype(x)
...
 @info "JacVec"
 
-L = JacVec(f, x)
+L = JacVec(f, x, 1.0, 1.0)
 @test L * x ≈ auto_jacvec(f, x, x)
 @test L * v ≈ auto_jacvec(f, x, v)
 @test mul!(dy, L, v) ≈ auto_jacvec(f, x, v)
 dy=rand(N);_dy=copy(dy);@test mul!(dy,L,v,a,b) ≈ a*auto_jacvec(f,x,v) + b*_dy
-update_coefficients!(L, v, nothing, 0.0)
+update_coefficients_for_test!(L, v, 3.0, 4.0)
```
vpuri3, line 87: Let's also add a `L(v, u, p, t)` test with different `p, t`

src/differentiation/vecjac_products.jl
```diff
@@ -65,26 +65,28 @@ struct RevModeAutoDiffVecProd{ad,iip,oop,F,U,C,V,V!} <: AbstractAutoDiffVecProd
...
 function update_coefficients!(L::RevModeAutoDiffVecProd, u, p, t)
+    update_coefficients!(L.f, u, p, t)
     copy!(L.u, u)
     L
 end
 
 # Interpret the call as df/du' * u
 function (L::RevModeAutoDiffVecProd)(v, p, t)
-    L.vecprod(_u -> L.f(_u, p, t), L.u, v)
+    L.vecprod(L.f, L.u, v)
```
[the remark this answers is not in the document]
vpuri3: if a user is passing in `u, p, t` then those values should be used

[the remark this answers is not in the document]
gaurav-arya: shouldn't it be consistent with the approach we took for `JacVec`? `L.f` depends on `p` and `t` implicitly.

test/update_coeffs_testutils.jl
```diff
@@ -0,0 +1,28 @@
...
+
+mutable struct WrapFunc{F,U,P,T}
+    func::F
+    u::U
+    p::P
+    t::T
+end
+
+(w::WrapFunc)(u) = sum(w.u) * w.p * w.t * w.func(u) 
+function (w::WrapFunc)(v, u)
```
gaurav-arya, line 11: @vpuri3 can you take a look at how this operator is using `u` as part of its state, and applying it to a different RHS `v`? Is this something that is invalid in your opinion? Because your recent changes don't work with this -- this also relates to the discussion in https://github.com/SciML/SciMLOperators.jl/issues/57.

src/differentiation/vecjac_products.jl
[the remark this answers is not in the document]
vpuri3: Yeah you are right. Let's be consistent bw JacVec and VecJac.

What we want to do nevertheless is to use `p, t` in the forward evaluation when provided with it. 

These two commits should ensure that

https://github.com/JuliaDiff/SparseDiffTools.jl/pull/232/commits/a182b8dfc8ff725aff0c134a1ff9f02e0141c2ee

https://github.com/JuliaDiff/SparseDiffTools.jl/pull/232/commits/b2bf791cb17bdbcb6efd6277246b0e00d7099cf0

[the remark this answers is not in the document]
gaurav-arya: (This is resolved)

test/update_coeffs_testutils.jl
gaurav-arya: (This is a separate, SciMLOperators issue, so resolved for our purposes here)

test/test_jaches_products.jl
gaurav-arya: (resolved)

gaurav-arya:
@ChrisRackauckas this PR's also ready for review

src/differentiation/jaches_products.jl
```diff
@@ -207,19 +207,23 @@ struct FwdModeAutoDiffVecProd{F,U,C,V,V!} <: AbstractAutoDiffVecProd
...
 end
 
 function update_coefficients!(L::FwdModeAutoDiffVecProd, u, p, t)
+    update_coefficients!(L.f, u, p, t)
     copy!(L.u, u)
     L
 end
 
 function (L::FwdModeAutoDiffVecProd)(v, p, t)
+    L = update_coefficients(L, v, p, t)
```
ChrisRackauckas, line 221: Is this correct? We want to do f'(u)*v, so the operator's values would be `u`, not `v`. We only need to update coefficients when `u` changes, as the operator itself shouldn't change when evaluating with a new `v`.

ChrisRackauckas: > I think my mental model of op(u,p,t) right now is:

But this is op(v,p,t) where the `v` is different from `u`.

> If it were just op * u, we wouldn't need the p and t parameters in the signature either, right, and just do op(u)? So if we're accepting the p and t here, it only makes sense if we're doing an update coefficients. For f'(u) v, we would do update_coefficients!(op, u, p, t) and then op * v. (And that's how e.g. ODE.jl should use the Jacobian, with the *)

`p` and `t` need to be there for if it's used as the ODE definition. In that case, `u' = Lu` with a JacVec operator would implicitly be solving `u' = f'(u)*u` for this reason. 

It might even make sense to just remove the `L(u,p,t)` operations in the near future as they weren't found to be all that useful due to this limitation.

gaurav-arya: Your original comment is resolved: lines are removed, and tests have been updated so that the would not pass if they were there.

gaurav-arya:
@ChrisRackauckas this is ready for another review. your comment is resolved (see above). There are a few cleanup points for a separate issue (https://github.com/JuliaDiff/SparseDiffTools.jl/issues/238), but the behaviour here should now be correct, and ready for ODE.jl.

Also: while making the tests more comprehensive I found that `x` and `v` ended up aliasing each other a lot in `test_jaches_products.jl` which resulted in the incorrect behaviour you caught not being picked up in tests. I fixed that, and also fixed an issue that `num_hesvec!` and friends did not restore `x` to its original state when mutating them, and then tightened up some unnecessarily loose tolerances that might have been in place because of this. All of this in done in c126bf8cb0, apologies for the additional diff noise: I needed these changes here in order to have comprehensive testing.

src/differentiation/jaches_products.jl
```diff
@@ -110,6 +111,7 @@ function numauto_hesvec!(dy,
     g(cache1, x)
     @. x -= 2ϵ * v
     g(cache2, x)
+    @. x += ϵ * v
```
gaurav-arya, line 114: @ChrisRackauckas that this was not before seems to actually have been a fairly serious numerical error: the vector x where the Jacobian is being evaluated at gets edited by around sqrt(eps) each time the Jacobian is applied, if I am thinking this through correctly

gaurav-arya: sorry, not Jacobian; this error was only present for (finite difference computed) Hessians

vpuri3:
@ChrisRackauckas ping

PR #244, opened 2023-05-30
Add downstream tests to ODE.jl, clean existing tests

vpuri3:

PR #245, opened 2023-05-30
Fix VecJac

vpuri3:
merge https://github.com/JuliaDiff/SparseDiffTools.jl/pull/244 first

In this PR, I have fixed some bugs in the implementation of `vector jacobian products`. In doing so, I have separated the `VecJac` implementations for `AutoFiniteDiff`, and `AutoZygote`. The latter lives in the `Zygote` extension. A concern with the Zygote VJP implementation raised in https://github.com/SciML/SciMLSensitivity.jl/pull/808 was that the underlying functions, `auto_vecjac(!)`, would recompute the pullback function every time the `FunctionOperator` is called by `*, mul!`. I have made it so that the pullback is recomputed only when `update_coefficients(!)` is called with the keyword argument `VJP_input`. 

```julia
L = VecJac(f, x1, p, t; autodiff = AutoZygote()) # pullback is computed at x1

# df/dx1' * v. pullback is not recomputed in below calls
L * v               
L(v, p, t)
mul!(w, L, v)
L(w, v, p, t)

L = update_coefficients(L, v, p, t) # pullback is not recomputed
update_coefficients!(L, v, p, t)    # pullback is not recomputed

update_coefficients!(L, w, p, t; VJP_input = x2) # pullback is recomputed at x2: Zygote.pullback(L.f, x2)

# df/dx2' * v. pullback is not recomputed in below calls
L * v               
L(v, p, t)
mul!(w, L, v)
L(w, v, p, t)

# df/dx2' * v --- pullback is recomputed at x2 Zygote.pullback(L.f, x2)
L(v, p, t; VJP_input = x2)
L(w, v, p, t; VJP_input = x2)
```

Tests are added in `test/test_vecjac_products.jl` to demonstrate this behaviour.


vpuri3:
It would be better, so as to avoid confusion between `u, v` in `df/du * v`, rename `u --> VJP_input` and pass it as a kwarg to `update_coefficients`. That way `L(v, p, t)` will be the same as `L * v`.

vpuri3:
This is good to go.  @ChrisRackauckas please take a look

vpuri3:
rerunning CI because of flaky ODE.jl "matrix contains infs/nans error"

ChrisRackauckas:
@avik-pal can you take a look? I know you looked at the vecjec potential usage in SciMLSensitivity and I think we should be trying to get it to there

vpuri3:
@avik-pal ping

avik-pal (approved):
This API looks great. We should be able to directly use it in SciMLSensitivty.jl after this is merged!

PR #250, opened 2023-06-15
Load Zygote in extension by checking for julia <1.9 support

### LINKED ISSUE #248: SparseDiffTools broken w/ Julia 1.8.5
The `Zygote` extension in the current version of `SparseDiffTools (v2.4.0)` cause `LoadError` when used with Julia 1.8.5:

MWE:

- `Project.toml`

> [deps]
Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
SparseDiffTools = "47a9eef4-7e08-11e9-0b38-333d64bd3804"

- `Loading`

> ┌ Warning: Error requiring `Zygote` from `SparseDiffTools`
│   exception =
│    LoadError: ArgumentError: Package SparseDiffTools does not have Zygote in its dependencies:
│    - You may have a partially installed environment. Try `Pkg.instantiate()`
│      to ensure all packages in the environment are installed.
│    - Or, if you have SparseDiffTools checked out for development and have
│      added Zygote as a dependency but haven't updated your primary
│      environment's manifest file, try `Pkg.resolve()`.
│    - Otherwise you may need to report an issue with SparseDiffTools
│    Stacktrace:
...
│    in expression starting at C:\Users\mstie\.julia\packages\SparseDiffTools\exnrN\ext\SparseDiffToolsZygote.jl:1
└ @ Requires C:\Users\mstie\.julia\packages\Requires\Z8rfN\src\require.jl:51

Seems the error is caused by the `import` statement
- https://github.com/JuliaDiff/SparseDiffTools.jl/blob/master/ext/SparseDiffToolsZygote.jl#3

Vaibhavdixit02:
Should fix #248

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -18,7 +18,7 @@ jobs:
         version:
           - '1' # Latest Release
           - '~1.6' # Current LTS
-          - '^1.9.0-0'
+          - '1.8'
     steps:
       - uses: actions/checkout@v3
       - uses: julia-actions/setup-julia@v1
--- a/Project.toml
+++ b/Project.toml
@@ -27,7 +27,7 @@ VertexSafeGraphs = "19fa3120-7c27-5ec5-8db8-b0b0aa330d6f"
 Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
 
 [extensions]
-SparseDiffToolsZygote = "Zygote"
+SparseDiffToolsZygoteExt = "Zygote"
 
 [compat]
 ADTypes = "0.1"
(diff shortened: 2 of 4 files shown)


.github/workflows/CI.yml
```diff
@@ -18,7 +18,7 @@ jobs:
         version:
           - '1' # Latest Release
           - '~1.6' # Current LTS
-          - '^1.9.0-0'
+          - '1.8'
```
Vaibhavdixit02, line 21: Do you want to keep the 1.9.0 here?

ChrisRackauckas: no. 1, 1.8, and 1.6 should be good.

PR #268, opened 2023-10-28
Fix Chunking

avik-pal:
- [x] Add tests
- [x] Add downstream testing

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -17,8 +17,7 @@ jobs:
           - InterfaceI
         version:
           - '1' # Latest Release
-          - '~1.6' # Current LTS
-          - '1.8'
+          - '1.6' # Current LTS
     steps:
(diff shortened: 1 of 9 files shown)


ChrisRackauckas:
```
[OOP] [Immutable AD]: Error During Test at /home/runner/work/SparseDiffTools.jl/SparseDiffTools.jl/downstream/test/basictests.jl:482
 Unexpected Pass
 Expression: abs.(ForwardDiff.derivative((p->begin
                    (benchmark_nlsolve_oop(quadratic_f, #= /home/runner/work/SparseDiffTools.jl/SparseDiffTools.jl/downstream/test/basictests.jl:483 =# @SVector([1.0, 1.0]), p)).u[end]
                end), p)) ≈ 1 / (2 * sqrt(p))
 Got correct result, please change to @test if no longer broken.

```

That sounds good.

PR #270, opened 2023-10-30
Fix tag

avik-pal:
@ChrisRackauckas this is the correct way to tag?


avik-pal:
Oh wait this is wrong, the previous version was correct https://github.com/JuliaDiff/ForwardDiff.jl/blob/d3002093beb88ff0b98ed178377961dfd55c1247/src/config.jl#L34-L43

PR #271, opened 2023-10-30
(Kind-of) Type Stability Fixes for  No Chunksize Specified

avik-pal:


avik-pal:
The ordinarydiffeq test failure is just some internet problem most likely. the nonlinearsolve ones will be handled by https://github.com/SciML/NonlinearSolve.jl/pull/265. Once the tests on that PR pass we can merge this

avik-pal:
@ChrisRackauckas this should be good to go

src/highlevel/forward_mode.jl
```diff
@@ -40,21 +45,23 @@ function sparse_jacobian_cache(ad::Union{AutoSparseForwardDiff, AutoForwardDiff}
...
 
 function sparse_jacobian!(J::AbstractMatrix, _, cache::ForwardDiffJacobianCache, f::F,
-    x) where {F}
+        x) where {F}
     if cache.cache isa ForwardColorJacCache
         forwarddiff_color_jacobian(J, f, x, cache.cache) # Use Sparse ForwardDiff
     else
-        ForwardDiff.jacobian!(J, f, x, cache.cache) # Don't try to exploit sparsity
+        # Disable tag checking since we set the tag to our custom tag
+        ForwardDiff.jacobian!(J, f, x, cache.cache, Val(false)) # Don't try to exploit sparsity
```
ChrisRackauckas, line 53: this is a nono

src/highlevel/forward_mode.jl
```diff
@@ -6,15 +6,20 @@ struct ForwardDiffJacobianCache{CO, CA, J, FX, X} <: AbstractMaybeSparseJacobian
...
+
 function sparse_jacobian_cache(ad::Union{AutoSparseForwardDiff, AutoForwardDiff},
-    sd::AbstractMaybeSparsityDetection, f::F, x; fx = nothing) where {F}
+        sd::AbstractMaybeSparsityDetection, f::F, x; fx = nothing) where {F}
     coloring_result = sd(ad, f, x)
     fx = fx === nothing ? similar(f(x)) : fx
     if coloring_result isa NoMatrixColoring
         cache = ForwardDiff.JacobianConfig(f, x, __chunksize(ad, x),
-            ifelse(ad.tag === nothing, SparseDiffToolsTag(), ad.tag))
+            __standard_tag(ad.tag, x))
```
ChrisRackauckas, line 22:
```suggestion
            typeof(__standard_tag(ad.tag, x)))
```

src/highlevel/forward_mode.jl
```diff
@@ -6,15 +6,20 @@ struct ForwardDiffJacobianCache{CO, CA, J, FX, X} <: AbstractMaybeSparseJacobian
     x::X
 end
 
+__getfield(c::ForwardDiffJacobianCache, ::Val{:jac_prototype}) = c.jac_prototype
+
 struct SparseDiffToolsTag end
 
+__standard_tag(::Nothing, x) = ForwardDiff.Tag(SparseDiffToolsTag(), eltype(x))
+__standard_tag(tag, _) = tag
```
ChrisRackauckas, line 14: How is that supposed to work? Do you wrap this in a ForwardDiff.Tag somewhere else? Seems like it's missing the wrap

ChrisRackauckas, line 14:
```suggestion
__standard_tag(tag, _) = ForwardDiff.Tag(tag, eltype(x))
```

avik-pal: I was assuming we wrap it while construction but it makes sense to do it here

src/highlevel/forward_mode.jl
avik-pal: But forwarddiff will check that the cache tag must match `Tag(f, v)`? How to deal with that

src/highlevel/forward_mode.jl
```diff
@@ -6,41 +6,53 @@ struct ForwardDiffJacobianCache{CO, CA, J, FX, X} <: AbstractMaybeSparseJacobian
...
 end
 
+__getfield(c::ForwardDiffJacobianCache, ::Val{:jac_prototype}) = c.jac_prototype
+
 struct SparseDiffToolsTag end
 
+function ForwardDiff.checktag(::Type{<:ForwardDiff.Tag{<:SparseDiffToolsTag, <:T}}, f::F,
+        x::AbstractArray{T}) where {T, F}
+    return true
+end
```
ChrisRackauckas, line 16:
```suggestion
```

This isn't necessary if done right

avik-pal: https://github.com/JuliaDiff/ForwardDiff.jl/blob/d3002093beb88ff0b98ed178377961dfd55c1247/src/config.jl#L42-L43 would trigger only if we don't wrap it into a Tag object

PR #272, opened 2023-11-01
Rework VecJac Operator

avik-pal:
- `p` and `t` are no longer dropped
- `VecJac` matches the size of $J^T$
- Non-square jacobian is constructed if we pass in `fu`

~I will test with https://github.com/SciML/NonlinearSolve.jl/pull/268 a bit before merging~

### DIFF
--- a/src/differentiation/vecjac_products.jl
+++ b/src/differentiation/vecjac_products.jl
@@ -1,10 +1,10 @@
-function num_vecjac!(du, f, x, v, cache1 = similar(v), cache2 = similar(v);
-        compute_f0 = true)
+function num_vecjac!(du, f::F, x, v, cache1 = similar(v), cache2 = similar(v);
+        compute_f0 = true) where {F}
     compute_f0 && (f(cache1, x))
     T = eltype(x)
     # Should it be min? max? mean?
     ϵ = sqrt(eps(real(T))) * max(one(real(T)), abs(norm(x)))
-    vv = reshape(v, size(x))
+    vv = reshape(v, size(cache1))
     for i in 1:length(x)
         x[i] += ϵ
         f(cache2, x)
@@ -14,9 +14,9 @@ function num_vecjac!(du, f, x, v, cache1 = similar(v), cache2 = similar(v);
     return du
 end
 
-function num_vecjac(f, x, v, f0 = nothing)
-    vv = reshape(v, axes(x))
-    f0 === nothing ? _f0 = f(x) : _f0 = f0
+function num_vecjac(f::F, x, v, f0 = nothing) where {F}
+    f0 === nothing ? (_f0 = f(x)) : (_f0 = f0)
+    vv = reshape(v, axes(_f0))
     T = eltype(x)
     # Should it be min? max? mean?
     ϵ = sqrt(eps(real(T))) * max(one(real(T)), abs(norm(x)))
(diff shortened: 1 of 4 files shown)


avik-pal:
This might have to be a breaking change. The previous API dropped things and a lot of code was written to work around it. But now the API is more strict which breaks code :disappointed:

src/differentiation/vecjac_products.jl
```diff
@@ -47,31 +51,121 @@ mul!(w, L, v) # = df/du * v
...
+f(u, p)     # p !== nothing
+f(u)        # Otherwise
+```
+
+For In Place Functions:
+
+```julia
+f(du, u, p, t)  # t !== nothing
+f(du, u, p)     # p !== nothing
+f(du, u)        # Otherwise
```
ChrisRackauckas, line 70: Does this match jacvec?

ChrisRackauckas: Where is v?

avik-pal: That also drops `p` and `t` after asking from them. JacVec basically only allows `f(du, u)` and `f(u)`. So this one is a strict superset

avik-pal: (I always pass in the general function and then get an error that `f` can't take `p` or `t` as inputs, which is kind of weird to do if the function takes them as inputs)

src/differentiation/vecjac_products.jl
```diff
@@ -14,9 +14,9 @@ function num_vecjac!(du, f, x, v, cache1 = similar(v), cache2 = similar(v);
     return du
 end
 
-function num_vecjac(f, x, v, f0 = nothing)
-    vv = reshape(v, axes(x))
-    f0 === nothing ? _f0 = f(x) : _f0 = f0
+function num_vecjac(f::F, x, v, f0 = nothing) where {F}
+    f0 === nothing ? (_f0 = f(x)) : (_f0 = f0)
```
prbzrg, line 18:
```suggestion
    _f0 = ifelse(isnothing(f0), f(x), f0)
```

avik-pal: This will cause an unnecessary computation of `f(x)` which is not needed

src/differentiation/vecjac_products.jl
avik-pal: > Where is v?

Oh I see the confusion. I meant to say function signatures for the input and not the constructed operator

src/differentiation/vecjac_products.jl
```diff
@@ -47,31 +51,121 @@ mul!(w, L, v) # = df/du * v
...
-    #       doesn't follow the convention of `f`
-    return FunctionOperator(L, u, u; isinplace = true, outofplace = OOP,
-        p, t, islinear = true, accepted_kwargs = (:VJP_input,), kwargs...)
+    fu === nothing && (fu = __internal_oop(ff) ? ff(u) : u)
+
+    op = _vecjac(ff, fu, u, autodiff)
+
+    # FIXME: FunctionOperator is terribly type unstable. It makes it `::Any`
+    # NOTE: We pass `p`, `t` to Function Operator but we always use the cached version from
+    #       VecJacFunctionWrapper
```
ChrisRackauckas, line 88: Sounds like the right thing to do is fix FunctionOperator?

avik-pal: the `note` part or the `fixme` part? The `fixme` part definitely needs fixing, but the `note` part is needed, because `p` and `t` need to be propagated via `update_coefficients`

PR #274, opened 2023-11-13
Fix type stability and allow JacVec to be non-square jacobians

avik-pal:


avik-pal:
@ChrisRackauckas the failure is because, we did not throw a warning before. The main change here is that if someone passes `p` and `t`, the general expectation should be that the function also accepts those as arguments. In the previous implementation, we dropped `p` and `t` after taking them as inputs, so I am giving a warning that the input function doesn't accept them as inputs

ChrisRackauckas:
That's fine, we should just make sure we fix downstream before merging so that way it's ready.

PR #276, opened 2023-11-29
Non allocating versions for StaticArrays

avik-pal:


src/highlevel/finite_diff.jl
```diff
@@ -10,6 +10,7 @@ __getfield(c::FiniteDiffJacobianCache, ::Val{:jac_prototype}) = c.jac_prototype
 
 function sparse_jacobian_cache(fd::Union{AutoSparseFiniteDiff, AutoFiniteDiff},
         sd::AbstractMaybeSparsityDetection, f::F, x; fx = nothing) where {F}
+    x = __make_mutable(x) # FiniteDiff is bad at handling immutables
```
ChrisRackauckas, line 13: how so? What is wrong with those dispatches?

avik-pal: I fixed it without this. Ig the assumption there is if `J` is mutable, `x` is implicitly assumed to be mutable in FiniteDiff, which is a reasonable assumption.

avik-pal:
The NonlinearSolve failures are fine here, it wasn't handling StaticArrays correctly in the first place, once the new PR is in, the failures should be patched.

PR #277, opened 2023-12-12
Path to use FiniteDiff for ApproximateSparsityDetection

avik-pal:


src/highlevel/coloring.jl
```diff
@@ -34,28 +34,70 @@ end
...
     J = fill!(similar(fx, length(fx), length(x)), 0)
+    J_cache = similar(J)
+    x_ = similar(x)
     for _ in 1:ntrials
-        x_ = similar(x)
         randn!(rng, x_)
-        J .+= abs.(ForwardDiff.jacobian(f!, fx, x_, cfg))
+        FiniteDiff.finite_difference_jacobian!(J_cache, f!, x_, cache)
+        @. J += abs(J_cache)
     end
```
ChrisRackauckas, line 101: Won't you want to threshold for numerical errors?

avik-pal: What would be a good threshold? `eps` is probably too small to be meaningful?

ChrisRackauckas (approved):

PR #282, opened 2024-01-16
Fix master

avik-pal:


src/differentiation/vecjac_products.jl
```diff
@@ -22,10 +22,11 @@ function num_vecjac(f::F, x, v, f0 = nothing) where {F}
     # Should it be min? max? mean?
     ϵ = sqrt(eps(real(T))) * max(one(real(T)), abs(norm(x)))
     du = similar(x)
-    cache = copy(x)
+    cache = similar(x)
+    copyto!(cache, x)
```
ChrisRackauckas, line 26: This dispatch isn't staticarray friendly anymore. We should make it use out of place `ArrayInterface.setindex`

avik-pal: I did the change because `copy(x)` might not be mutable but `similar(x)` is supposed to create a mutable version, right?

ChrisRackauckas: That's why I'm saying we should use the out of place `setindex` instead of `Base.setindex!` so that it doesn't require mutation.

avik-pal: Ah okay. Let me do that.

ChrisRackauckas: allowed_setindex so it supports GPU

avik-pal: `setindex` will still trigger a scalar indexing in that case right?

PR #290, opened 2024-04-16
Don't use custom tags

avik-pal:
I don't remember why we used them in the first place, but it causes tag ordering issues for higher order AD.


avik-pal:
cc @ChrisRackauckas

ChrisRackauckas (approved):

avik-pal:
@ChrisRackauckas you need to merge.

PR #291, opened 2024-04-18
Add Enzyme v0.12 compat

gdalle:
This is blocking DifferentiationInterface tests where I wanna use the latest version of Enzyme.
Hopefully the test workflow does pick Enzyme v0.12, otherwise we might have to set `force_latest_compatible_version = true` in the `.yml`


gdalle:
The integration test [already fails on `main`](https://github.com/JuliaDiff/SparseDiffTools.jl/actions/runs/8741946402/job/23989109431)

PR #296, opened 2024-05-02
Re-add Julia 1.6 compat for the time being

gdalle:
Follow up to https://github.com/SciML/ADTypes.jl/pull/52


gdalle:
Blocked by PolyesterForwardDiff, which dropped Julia 1.6 compat in their v0.1.1 release

PR #302, opened 2024-10-07
Set default `rows_index` and `cols_index` to `nothing`

gdalle:
Partial fix for https://github.com/SciML/NonlinearSolve.jl/issues/473

A better solution would be something like this
```julia
julia> using Base.Iterators

julia> using BenchmarkTools

julia> function findstructralnz(x::AbstractMatrix)
           cartind = vec(CartesianIndices(x))
           return Iterators.map(first ∘ Tuple, cartind), Iterators.map(last ∘ Tuple, cartind)
       end
findstructralnz (generic function with 1 method)

julia> collect.(findstructralnz(rand(2, 3)))
([1, 2, 1, 2, 1, 2], [1, 1, 2, 2, 3, 3])

julia> @btime findstructralnz($(rand(2, 3)));
  28.162 ns (0 allocations: 0 bytes)
```

But for some reason we don't have code coverage information for this repo, so I'm not even sure this branch is getting hit in tests. Better err on the side of caution until that changes.

### DIFF
--- a/src/differentiation/compute_jacobian_ad.jl
+++ b/src/differentiation/compute_jacobian_ad.jl
@@ -348,8 +348,8 @@ function forwarddiff_color_jacobian!(J::AbstractMatrix{<:Number},
     if FiniteDiff._use_findstructralnz(sparsity)
         rows_index, cols_index = ArrayInterface.findstructralnz(sparsity)
     else
-        rows_index = 1:size(J, 1)
-        cols_index = 1:size(J, 2)
+        rows_index = nothing
+        cols_index = nothing
     end
 
     # fast path if J and sparsity are both AbstractSparseMatrix and have the same sparsity pattern


ChrisRackauckas:
This isn't quite correct either? It's correct for Matrix, but other structured sparse matrices like banded matrices do still have a non-trivial sparsity pattern, but don't use structurednz. However, in that case these are just sentinals, so I guess it's fine to have this, especially since this is legacy code at this point that should get archived in the very near future.

ChrisRackauckas (approved):

gdalle:
> It's correct for Matrix, but other structured sparse matrices like banded matrices do still have a non-trivial sparsity pattern, but don't use structurednz.

The previous version was incorrect too, it basically assumed the matrix was diagonal.
At least with this version, every pair of indices is listed, which means we'll get a cleaner error with other structured matrices. But I would also like to remove the `@inbounds` in this package and in the FiniteDiff extension, to prevent bad surprises.

gdalle:
For testing this it would be great to have code coverage (#303), any idea why we don't?

ChrisRackauckas:
No idea what happened to code coverage. That's news to me that it's not working. Though it has always been a bit wonky with test groups.

ChrisRackauckas:
> The previous version was incorrect too, it basically assumed the matrix was diagonal

Agreed, that's why I still approved it. It's a sentinel in the other cases, so it should be fine? Wanted to make sure that was documented in case anyone looked back at this PR as to why it's not the right value.

gdalle:
Alternately we could put `nothing` as you had suggested earlier. Without code coverage, it's hard to know whether this branch is even hit

gdalle:
Fixed in https://github.com/JuliaDiff/SparseDiffTools.jl/pull/305

ChrisRackauckas:
Looks like this breaks GPU

gdalle:
https://github.com/JuliaDiff/SparseDiffTools.jl/issues/307

gdalle:
What do you want to do here @ChrisRackauckas?
We somehow have to generate two sequences that look like `[1, 2, 3, 1, 2, 3]` and `[1, 1, 2, 2, 3, 3]`. If we use lazy iterables, GPU is not happy, but if we allocate, well we allocate.

gdalle:
Gentle bump

Byurdyzo (approved):

Byurdyzo (approved):

PR #304, opened 2024-10-07
Remove or sanitize inbounds

gdalle:
Remove some `@inbounds` which were used in an unsafe way.


ChrisRackauckas (approved):

ChrisRackauckas:
Can you fix that downstream test lts version?

gdalle:
What do you mean?

gdalle:
Oh right, it was a quoting error, see https://github.com/JuliaDiff/SparseDiffTools.jl/pull/306

ChrisRackauckas:
it's downstream testing 1.1 because it's not "1.10"

